PR #9, opened 2022-09-27
D20127858 joey corcoran/branch1

joeycor92:
Hi Folks, 

Here's mine. 

The only change was an addition to the top of the readme file where I added in team members and my name.

image
Also testing to see whether pictures send as if we add front end features it's good to show how they look for reference.

Thanks
Joey


DaraghK93:
Looks good. Picture sent as well

PR #244, opened 2022-10-03
added bootstrap configuration to project and test navbar

beckskelly:


beckskelly:
image

joeycor92 (approved):
As discussed, not sure if the header change should be here right now as you're still working on it, but it's ok for the moment. The Bootstrap config looks good to me and once Warren reviews we can merge.

WarrenK-design (changes requested):
This was changed not sure if that is correct, think it was fine before from Joeys stuff 
image

beckskelly:
Good spot on imports, modified back to original and merging to origin now

PR #251, opened 2022-10-03
Added Mongoose Schemas for user.model stock.model

BearachB:
…in backend. user.test.js needs to be revisited and user model may need to be updated at a later date.


caolan-power (approved):
Nice one. All looks good

PR #252, opened 2022-10-03
Redux store creation

WarrenK-design:
The redux store has been created finishing the issue Create Redux store with this commit. Work has started on the user reducer and actions for login however a new branch and issue will be opened to finalise this functionality.


joeycor92 (approved):
Thanks for this. From what I can tell everything is up to date. Just to note that I only looked at the first commit here as you mentioned that the second was just from pulling from master to get your branch up to date.

WarrenK-design:
@DaraghK93 you might have quick glance at second commit the backend stuff, was just a merge from master so should be ok but just to make sure

beckskelly (approved):
Looks good, thanks for setting this up!

DaraghK93 (approved):
looks good. The backend changes are just the removal of the two original Lambda functions and the addition of our new expressServer lambda function.

PR #256, opened 2022-10-03
Mvp backend userroute

DaraghK93:
changed userRoutes, to show the three routes so far. userController just has a simple json response when there's a request sent. cleaned up app.js also


caolan-power (approved):
I like the structure here

PR #262, opened 2022-10-04
Added portfolio model schema to models directory

BearachB:


caolan-power (approved):
All good. Might cause a conflict with my changes. Will check now

PR #263, opened 2022-10-04
First stock routes

caolan-power:
First Stock routes implemented. Getting creating and updating. 

I commented out some of the db schema for now until we flesh out what we need so it's reduced to name and price.

Also, I might change name to symbol or ticker or whatever we decide but 'name' for now will do.


caolan-power:
using mongoose.find() for now. I'm not sure is this bad practice. findOne() was throwing an error for some reason. It returns an array so that's why I need to say stock[0].price etc

BearachB (commented):
Everything looks good to me. I wonder there a better way of handling the model schema other than commenting it out? None that I can think of right now but just wondering in case we have this problem again

DaraghK93 (approved):
looks good

PR #265, opened 2022-10-05
Backend user details

DaraghK93:
changed login and registration on the backend so that all the details, bar password, are sent in the response


WarrenK-design (changes requested):
For the returning JSON always make it key value pairs as I need to know in frontend what keys to use to access the details, at the moment if I tried to access them they are all just in one object would just look like {warren, warren@gmail.com, warren123, jwtToken} I'm pretty sure. 

Also for register screen @joeycor92  is sending the first name and the lastname, "name" should be broken up on backend too and in the database @BearachB might need to change schema then. There is no username either in the register screen so we need to all have a look at this and agree what is being sent.  
image

The return JSON should look like this below then. 
res.json({
  userID: <valueOF_ID_FROM_MONGO>,
  firstName: <value>,
  secondName: <value>,
  username: <Value>,
  email: <Value>, 
})
For these too never send back the password to the user in the response. 
image
image

BearachB:
OK, will discuss what user info we need to store in the stand-up today and update the schema for it.

DaraghK93:
> For the returning JSON always make it key value pairs as I need to know in frontend what keys to use to access the details, at the moment if I tried to access them they are all just in one object would just look like {warren, [warren@gmail.com](mailto:warren@gmail.com), warren123, jwtToken} I'm pretty sure.
> 
> Also for register screen @joeycor92 is sending the first name and the lastname, "name" should be broken up on backend too and in the database @BearachB might need to change schema then. There is no username either in the register screen so we need to all have a look at this and agree what is being sent. image
> 
> The return JSON should look like this below then. res.json({ userID: <valueOF_ID_FROM_MONGO>, firstName: , secondName: , username: , email: , }) For these too never send back the password to the user in the response. image image

Was just discussing this with Caolan. as the response object has been restructured we don't need to explicitly call the key value pairs. can just say "name" and it treats it as "name:name", if that makes sense. 

I've taken out password as well, that was an oversight

WarrenK-design (changes requested):
image
This is fine if it does the key values but again it should link up with Bearach schema, he has first name, secondname and DOB too.

joeycor92 (approved):
Following our chats and the call this morning I agree with these changes once Warren's request has been implemented. We only need first name to be returned really? Second name and DOB won't be used anywhere else will they?

WarrenK-design:
> Following our chats and the call this morning I agree with these changes once Warren's request has been implemented. We only need first name to be returned really? Second name and DOB won't be used anywhere else will they?

We will need them for the profile screen/edit details yeah, its best off do this now once.

joeycor92:
> > Following our chats and the call this morning I agree with these changes once Warren's request has been implemented. We only need first name to be returned really? Second name and DOB won't be used anywhere else will they?
> 
> We will need them for the profile screen/edit details yeah, its best off do this now once.

Oh yes that makes sense. Thanks

BearachB (approved):
Schema has been updated to reflect this.

WarrenK-design (approved):
Looks good cheers for update Daragh

PR #268, opened 2022-10-05
229 user schema

BearachB:
updated user schema to capture all frontend account creation fields. Formatted the other schemas too to make them more readable.


DaraghK93 (approved):
looks good. we will probably save the python script in a separate web scraping folder eventually though

joeycor92 (approved):
Looks good for a start. As mentioned will likely change over time but good for now.

PR #273, opened 2022-10-05
Login redux

WarrenK-design:
User is now able to login through the frontend and their details are saved in redux. The error message still needs to be saved in redux also when agreed upon format is finalised with backend. 

For frontend I've changed the API route but its in the aws-exports.js file under src which doesn't push to GitHub, looks like the first picture. I've added the "localAPI" field which its destination is set to localhost:3000 where the express API runs on your machine. In the file constants/APIconstants.js there is the line shown in the second picture which detects if you are running react in dev mode and if you are it will call API on your local machine which you are working on instead of going to the production API on the master branch.

image
image


joeycor92 (approved):
From what I can tell everything is correct there. As mentioned, if you want to throw in the redirect you should do so and push again.

beckskelly:
Looks good to me!

beckskelly (approved):

WarrenK-design:
> From what I can tell everything is correct there. As mentioned, if you want to throw in the redirect you should do so and push again.

Added redirect there only in login page, will be added in becks push for App.js route

joeycor92 (approved):
Great thanks for making that change

DaraghK93 (approved):
yeah, all seems good.

PR #274, opened 2022-10-05
Removed the required field for each symbol in portfolio model to make…

BearachB:
… testing easier


DaraghK93 (approved):
yes this makes more sense, so we're not storing stocks the user doesn't have

PR #275, opened 2022-10-05
271 login with google button

joeycor92:
As mentioned, the colour may change in the future but for now we have the below. Also included is a component for the button which can be included on the registration page.
image


beckskelly (commented):
Same review as @WarrenK-design , don't hard code the form action, also where did you find the icon?

PR #276, opened 2022-10-05
Create error middleware

caolan-power:


DaraghK93 (approved):
thanks for updating those. looks a lot better

PR #277, opened 2022-10-05
Register redux

WarrenK-design:
Has the functionality for the action and reducer for registering a user and saving the new users state to redux. Still need a register page to test this functionality. Small fix for backend "next undefined" error and errormessage undefined error.


DaraghK93 (approved):
thanks for fixing the next error. seems good to me

PR #281, opened 2022-10-06
Component created called MessageAlert

WarrenK-design:
This can be used for warning error, info or any other kind of messages. It uses bootstrap variants (https://bootstrap-vue.org/docs/reference/color-variants) for now, this may need to be changed in future to our own colour scheme. Comments in the file MessageAlert.js explain how to use it and there is an example using the error stored in redux for bad login request. 

image


beckskelly (approved):
Looks good, well explained in the comments!

PR #282, opened 2022-10-06
248 registration form

joeycor92:
I've added the boxes for the registration form as well as an error message for Password and Confirm password. More will need to be added. Also everyone will need to do 
`npm i`

Check that this lines up with the register user action.
image


WarrenK-design (approved):
Looks good!

PR #288, opened 2022-10-06
Revert "248 registration form"

joeycor92:
Following the spot from Warren that there was something amiss with Username, I revisited the code and noted that this was not the only issue as the type of password should also have been "password".

All inputs are also now required before the user can proceed.


joeycor92:
Revert seemed to delete the changes altogether. As such I just deleted the branch and pushed again.

PR #293, opened 2022-10-06
Redux registration page

WarrenK-design:
Redux now working with registration page. Makes POST request to /api/user in the register user action, then logs the user in using the login redux action. 

The state is saved in user -> userInfo in redux and contains info below. 

image


DaraghK93 (approved):
looks good. cheers for the clean up. I'll add that validation for nulls

joeycor92 (changes requested):
Looks good except maybe take out the console.log
image

WarrenK-design:
> Looks good except maybe take out the console.log image

Updated there

joeycor92 (approved):
Thanks for that change.

PR #295, opened 2022-10-06
Display logged in user

WarrenK-design:
Updated Nav bar, added in few more links they can be changed later and now logged in user displays in the top right. 

image
image


joeycor92 (approved):
Looks good to me and looks great on the page. Thanks. Will merge

PR #296, opened 2022-10-06
added stockdisvoery page

beckskelly:
Added stock discovery page,
added components for searchbar (not functional) and tickercard
Fixed footer (position relative)
Pulled in stocks from DB to populate tickercards


beckskelly:
image

WarrenK-design (changes requested):
image
Cant see it in your package.json either?

WarrenK-design (changes requested):
**This is formatted as email** 
image

*****************************************************************************************

**Are we fixing this spacing?** 
image

******************************************************************************************

**Mobile one too**
image

joeycor92 (changes requested):
As mentioned in the messages you might want to change this:
image

We shouldn't be hardcoding the url.

As well, not for now but we need to discuss whether we're using axios or async so that we can be consistent.

WarrenK-design (approved):
Fixed CSS and API now calls the live API not the localhost when deployed.

joeycor92 (approved):
This looks great. Thanks @beckskelly for doing the brunt of it and thanks Warren for helping with the CSS and a few bits.

PR #297, opened 2022-10-06
Logout feature

WarrenK-design:
Have added the logout action to userActions. When a user is logged in the option to logout will appear in the nav bar. 

The nav bar has also been updated to have few more links for profile, stocks, login, register, logout. 

Styling of navbar has been changed. 

New component called "FormContainer.js" created, can use it for all forms from now on the get same styling. 

image

### DIFF
--- a/src/components/layout/Header/header.js
+++ b/src/components/layout/Header/header.js
@@ -36,29 +36,27 @@ function Header() {
   }
 
 
-
-
   return (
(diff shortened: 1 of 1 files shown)


joeycor92 (approved):
Brilliant, logout ready to go so. FormContainer does a much better job as well

beckskelly (approved):
Looks good to me!

PR #300, opened 2022-10-10
Twitter schema and routes

caolan-power:
Added twitter model, basic routes and controllers. Need this so I can test my lambda function for twitter scraping


DaraghK93 (approved):
looks good. I'm wondering do we even need a post route if the tweets will just be coming straight from the API?

BearachB (approved):
Not sure if it's entirely necessary for this one, but in Mongoose the "trim : true" option will remove whitespace from the start and end of DB entries into Mongo. Might be used down the line when making new/updating schemas

caolan-power:
> Not sure if it's entirely necessary for this one, but in Mongoose the "trim : true" option will remove whitespace from the start and end of DB entries into Mongo. Might be used down the line when making new/updating schemas

Will update to include this. Probably good to have. Thanks

PR #301, opened 2022-10-10
Tweet scraping lambda function

caolan-power:
Creating this pull request to get more eyes on this problem.

The lambda function works as expected however the mongouri is in the code. 

I have added a secret from the cli called mongourl and i try to call it in the code but i am getting a permissions error.

I pushed this to amplify to test using the amplify console so you can see it there too if needed.


WarrenK-design (changes requested):
This function is nested in handler, should be its own function separate to handler. 

image

Im not sure on this part, could you try the tutorial I sent on the Amplify docs they give different code there and how to add secrets through CLI https://docs.amplify.aws/cli/function/secrets/#accessing-the-values-in-your-function 
image

Can you add an image of permission error you are getting?

caolan-power:
> This function is nested in handler, should be its own function separate to handler.
> 

Well spotted. I moved it in there getting around some error I was getting. Will move it back now

> image
> 
> Im not sure on this part, could you try the tutorial I sent on the Amplify docs they give different code there and how to add secrets through CLI https://docs.amplify.aws/cli/function/secrets/#accessing-the-values-in-your-function image
> 

Added through the CLI yes but when pushed still permissions error even after configuring with Daragh. 

> Can you add an image of permission error you are getting?

Error is 'MongoURL because no identity-based policy allows the secretsmanager:GetSecretValue action'

However now Im getting a resource not found error. Will organise a meeting to troubleshoot

WarrenK-design:
Ok try this code for getting secret seems to be working on my local setup. Updated your permissions in AWS and created a single secret called ```Mongo_URI``` that all functions/express can use. 

Do a ```amplify update function``` and remove the secret you created for that function too, I have deleted it in the AWS console.

```
def getSecret(secretName,region="eu-north-1"):
    """
    Description:
        This function is used to get a secret from the parameter store in AWS. 

    Args:
        secretName (string): The name of the secret found in the parameter store. 
        region (string): The region the secret is stored in, default to eu-north-1. 

    Returns:
        response (dict): Dictonary object with the secret name under "Name" key and value under "Value" key
    """ 
    client = boto3.client('ssm',region)
    try:
        response = client.get_parameter(
            Name=secretName,
            WithDecryption=True
        )
        return response
    except Exception as e:
        print(f'ERROR:Could not get secret in getSecret function.\nException Details:\n\t{e}')

mongoURI = getSecret('MONGO_URI')
```

WarrenK-design (changes requested):
Should write the code and test it now anyway. You are definitely going to have to write code by Friday anyway, think it should be finished out, we aren't just presenting just Apple on Friday, I don't think. Even locally using your own Mongo setup. 
image

caolan-power:
Can do. Was thinking just to keep free tier usage down but I guess we've passed that already. Will update now to pull 2 tweets from each of the 500 companies

WarrenK-design (changes requested):
Just small things like this if copying code. Have a look 

image

Have a look at Python docstings here they are really useful in Python https://peps.python.org/pep-0257/. 

For VSCode download this package, once you write a function you can then generate the docstrings using it will make code a bit clearer. 
image

Also all the environment variables you use have also been added through amplify cli yeah?
This limit one may be good use for environment variable too so we can easily change it later instead of editing code, say we want 10 tweets in future. 
image

caolan-power:
Ok cleaned up the comments and changed limit to an env variable.

WarrenK-design (approved):
Looks good!

joeycor92 (approved):
Looks good to me! Don't think I have any changes. I really just checked index.py and the cron function as everything else should be autogenerated.

PR #302, opened 2022-10-10
Issue #260 Set up SCSS

beckskelly:
Set up SCSS. At the top of the file you can see the bootsrap is pulled in and alterations are made  below it. Some variables were set and assinged to the scss. For now I've kept it all in one file as we don't have a lot of css but in the future we can break it out to different files if need be.

You can see on the .btn-primary class and its sub classes that I've used a function to darken it by 10% on clicked/ hover, and had to use the !important tag to override the bootstrap css. I also nested the h1 class inside .navbar brand, changing only h1's inside that. Hope this shows some of the functionality of scss!  

image


WarrenK-design (approved):
Looks good becks, again I personally like white background but that's only small think will go with majority/user evaluation maybe, good work.

PR #304, opened 2022-10-11
Added script that was used to generate and push the image urls from S3…

BearachB:
…to Mongo. Also moved all Python scripts to a new top level directory in project folder


BearachB:
One thing I've just realised is that the Mongo URI and username/password are exposed in this. Will figure it out and recommit in a minute with a solution

DaraghK93 (approved):
looks great. thanks for that

PR #305, opened 2022-10-11
News articles scraper Lambda function

WarrenK-design:
The function at the moment scrapes news sources only and stores them in the database. The sentiment is hardcoded for this week  as the training of the sentiment classifier will be done next week. 

There is a README file located [here](https://github.com/DaraghK93/stockApp/blob/newsArticlesScraperLambdaFunction/amplify/backend/function/newsArticleScrapingAndSentimentAnalysis/README.md) if you want to run the script locally you need to create a .env also locally. 

The schedule is set to run every 30 minutes which will be created once this branch is merged to master. 

Articles can be viewed in the ```articles``` document of the database. 

Articles will only be scraped if they contain title, description, link, publication date and it has been published since the last scrape has happened.  

If an article does not have an image associated with it, it will be assigned a default image being the news sources thumbnail. 

For adding new news sources the file in src/lib/RSSSources.json needs to be updated as in the README.md file.


caolan-power (approved):
Really cool. I need to use the .env you use here in my own for the db name. 

Also, I like the BulkWriteError and I'm going to steal it for my function because I'm doing a naive approach coded up myself. 

Looks good

joeycor92 (approved):
Bit of reading in that! Looks fantastic, might need to chat to you about the crontime later for my webscraping. Commenting is really helpful as well.

DaraghK93 (approved):
looks great. I have to make note of your exception handling as well for mine

PR #306, opened 2022-10-12
Delete stock function

DaraghK93:
deleting the stockFunction folder as it is old and not used. it has been removed from amplify. I don't know why it showed up again but it was causing some error in AWS so best to delete it for good.


BearachB (approved):
That's fine, I'd be curious why it showed up again if anyone figures it out.

WarrenK-design (approved):
Yeah I noticed this too. Approved but it does say only package.json edited which is strange 

image

PR #308, opened 2022-10-12
Error for directory not being found on Lambda and other errors 

WarrenK-design:
Think this is fixed unsure until it is run on Lambda, may be trial and error process of fixing these small errors with lambda at the start unless anyone has better way to speed up process. 

image


caolan-power:
I had to make those changes on my local machine too. Hopefully will line up well when deployed then

caolan-power (approved):

DaraghK93 (approved):
happy for you to do it by trial and error. once we have it sorted for one anyway we should be ok for the rest of them

PR #309, opened 2022-10-12
300 get historical data

joeycor92:
This is the python file that was used to populate the historical prices. This takes data from the Alpha Vantage API. This has limits of 500 requests per day + 5 requests per minute so this had to be taken into account. As such there is a timer which runs a function every 60 seconds.

As well as this, because of the huge amount of data that was collected, the stock_controller.js file had to be changed to not get all of the prices that were put into the database.


BearachB (approved):
Looks good, great work Joey! Thanks for updating the stock controller too. Just one thing to note, we might need to have a think about the structure of the database directory and try lay it out in a way that makes sense and also reduces the need for redundant csv and Python files.

BearachB:
> I think you need to add a few exception handlers to the main script. such as when you're requesting from the API and, connecting to the DB and sending the bulk write as well. If you look at Warren's he has them everywhere. if there's an error when it runs, we'll atomatically see "database connection error" or something and will save a lot of time.

Are we going to need to run this script again though? Once it's populated the DB with the prices we won't need it again, as we'll be scraping the prices usuing Beautiful Soup

BearachB:
> Are we going to need to run this script again though? Once it's populated the DB with the prices we won't need it again, as we'll be scraping the prices usuing Beautiful Soup

But yes ultimately Daragh is correct, should add them in. I didn't properly examine the Python script as I was thinking we weren't going to need it again.

joeycor92:
> I think you need to add a few exception handlers to the main script. such as when you're requesting from the API and, connecting to the DB and sending the bulk write as well. If you look at Warren's he has them everywhere. if there's an error when it runs, we'll atomatically see "database connection error" or something and will save a lot of time.
> 
> also: Screenshot 2022-10-12 at 12 25 21
> 
> Screenshot 2022-10-12 at 12 25 40
> 
> small thing, but you have comments here that don't relate to these files.
> 
> Screenshot 2022-10-12 at 12 26 19
> 
> and maybe take this print statement out? I assume was just for debugging

Thanks both @DaraghK93 @BearachB for your comments. Completely agree, and it was an oversight to exclude these try/excepts. This may need to be run again so important to have these checks.

DaraghK93 (approved):
great, thanks for changing Joey

BearachB:
Very good spot Daragh!

PR #316, opened 2022-10-12
Individual stock page and stock visualisations started

beckskelly:
Created individual stock page with route from /stockdiscovery page that Warren made last week.
Created components for graph visualisations - pie chart, radar chart, bar chart. Dummy data used
Created price chart component that changes dummy data based on button clicked
Created 2x card components, one for changing charts, and one plain card. Have it so that certain info will auto display a certain chart first.
Created information button and modal pops up when clicked for the stock information card at top of page. none of the others function right now.

image


joeycor92 (approved):
Looks good for now. One thing we might need to look at is the size of the logo, I think that this needs to be smaller and in the top left corner maybe. Something to look at in the future as I know this is a first draft. It just seems absolutely massive the way it is an takes up most of the page.
image

Thanks as well for demonstrating this to us as well. The graphs look great and they're doing exactly what we want. Not too sure if I'm all in on the colours yet but that's easily changed.

The main note I have is that it's very difficult to check so much code at once, I found myself going back and forth between lots of different files to check stuff. We should be doing these pull requests in smaller increments, e.g. per chart, so it's easier to spot bugs.

beckskelly:
@joeycor92 agreed on size of code, will not do that again, if you want I can demo it to you guys and explain everything after the standup

beckskelly:
@joeycor92 also good point on the logo, probably lots of other bits that are sub-optimal right now, after this PR maybe we could write a list of stuff to fix and sort it out before Friday

joeycor92 (changes requested):
Sorry I had issues with the laptop and I was looking at an old example when it compiled. It just refreshed and can now see there are a good few errors showing up.
image

These are the whole way through so I'd say make sure to fix.

WarrenK-design (changes requested):
Hi Becks looks really good but few things just listed below. 

As well I think its a bit unfair asking myself and Joey or whoever to review this large pull request and waiting until a Thursday to do it, the same happened last week with your stock discovery pull request was a lot to review. 

This could have been broken up through out the week as in a PR for the logo at top, bar chart, pie chart, radar chart etc. There is a lot of code and below is only some of things I caught.  

**1. Button Spacing should be equal** 
image
**2. Logo at top**
Same as Joeys comment this just looks out of place. 
image
**3. Graph responsiveness** 
When you change screen size weird things happen.
image
**4. White Space**
image
**5. Graph Labels**
image
image
**6. CSS not Used**
I taught we where going with the scss you set up?
image
**7. Logic Wrong**
image
**8.Chart Type**
Should just have chart type as an input to the prop, makes it more reusable. 
image
**9. Remove**
Remove these comments. 
image
**10. Log Errors**
Should be none of these. 
image

joeycor92 (changes requested):
Thanks for making those changes. No errors showing up now in terminal. Although I am getting these errors here
image

I think we need to reshuffle the logo and prices (going against the prototype here) and possibly move these to the right side of the logo on mobile. This might eliminate the whitespace.
image

I assume you've concentrated on mobile first, because a lot will probably need to be done regarding the layout for desktop.

Not requesting changes so much as wondering about the errors.

beckskelly:
Hey @joeycor92 I've noticed that weird error, I think its something to do with Responsive container in the recharts package, I think when you make the console window smaller than a certain size it throws that error. I'll check it out after, I'm nearly done all the changes @WarrenK-design suggested. Agreed on the giant pull request being bad form so sorry about that guys, will take it on for next week to stop

WarrenK-design (changes requested):
**1.White Space**
I still think this looks very squished on mobile. Still space white space just with text in it. font seems different too. 
image
**2.Mobile Graphs**
image
image
**3.Desktop View**
Just spacing here you can fix same way stock cards fixed cards last week. 
image
**4.Logo**
Maybe try aim for something like the prototype again just looks bit out of place. 
image

image

joeycor92 (changes requested):
Think the white space is really the main issue, if we can reduce that we'd be flying it.

beckskelly:
Yeah ill have to do some messing around, It seems to want to put that space there regardless of there being a label there or not, see screenshot

image

joeycor92 (approved):
As mentioned by Warren, happy for these to be pushed through. A few bits left to fix but good for now and it'll be good to show them off tomorrow! Thanks for this.

PR #328, opened 2022-10-13
Esg data ingress change env var

DaraghK93:
changed environment variable for DATABASE so it would update our live main "test" DB


WarrenK-design (approved):
Approved just make sure with joey we still working off sample_stock_data collection for now

caolan-power (approved):
looks good

PR #332, opened 2022-10-13
Search article api route

WarrenK-design:
Route for GET /api/newsarticles created. If this is called with no query params then all articles returned. If it is called with the query params ```symbol```, ```shortname``` and ```longname``` related to the stocks MongoDB data then it will get news related to the given stock. 

Services directory also created in the backend for reusable functions and calling models as per this architecture https://www.coreycleary.me/project-structure-for-an-express-rest-api-when-there-is-no-standard-way/ means we don't have to write things like searching functions for each article route. 

The searching wors for the most part but could be improved with keywords list for companies.


DaraghK93 (approved):
looks great. thanks for creating the services folder. will definitely be useful

caolan-power (approved):
Cool functionality. Can use something similar for searching anywhere in the app now

PR #333, opened 2022-10-13
User details validation

DaraghK93:
put in validation on the backend for user registration and login. 400 error if they do not fill all fields.
Screenshot 2022-10-13 at 18 26 53

Also fixed one error field which should have been 404
Screenshot 2022-10-13 at 18 27 17


amplify/team-provider-info.json
```diff
@@ -15,25 +15,29 @@
       "function": {
         "expressServer": {
           "deploymentBucketName": "amplify-stockapp-dev-151619-deployment",
-          "s3Key": "amplify-builds/expressServer-746764634d6933773954-build.zip"
+          "s3Key": "amplify-builds/expressServer-753174514c376f647362-build.zip"
```
caolan-power, line 18: What are these changes? Other than being unsure of these everything looks good to me

caolan-power (approved):

amplify/team-provider-info.json
```diff
@@ -15,25 +15,29 @@
...
+          "s3Key": "amplify-builds/twitterScraping-395a4c30485a69633149-build.zip",
           "databasename": "test",
           "limit": "3"
         },
         "newsArticleScrapingAndSentimentAnalysis": {
           "secretsPathAmplifyAppId": "d2l3aeky8jssr5",
           "environment": "prod",
           "databasename": "test",
-          "crontime": "30"
+          "crontime": "30",
```
BearachB, line 31: The changes in this file are separate to the user validation? Not sure how this fits in if you could give a small explanation

BearachB: Sorry I'm half asleep here, now I see that this is the same problem from earlier. All good, but we need to figure out what's going on with it for future reference

BearachB (approved):
Just one comment on team-provider-info.

PR #334, opened 2022-10-13
Created the individual news cards

WarrenK-design:
They change oriented by row or col based upon phone or desktop. New sentiment badge component also added.
Image size is issue used tricky image in example below, on phone to keep ratio this is what looks like, what do we think? Was trying for figma design but the text in figma was actually tiny when used that CSS. 

Note dates be fixed in higher component which data passed from. 
***Desktop***
image
***Phone - Iphone SE***
image
***Prototype***
image

### DIFF
--- a/src/components/newsComponents/newsArticleCard/NewsArticleCard.js
+++ b/src/components/newsComponents/newsArticleCard/NewsArticleCard.js
@@ -1,0 +1,39 @@
+import {Card} from 'react-bootstrap'
+
+import SentimentBadge from '../../widgets/sentimentBadge/SentimentBadge';
+
+
+/// NewsArticleCard ///
+// Description:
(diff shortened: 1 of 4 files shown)


beckskelly:
You could get away with making them smaller, see snip from the business times 
image

WarrenK-design:
New mobile cards
image

beckskelly (approved):
Looks good to me! I like the smaller card better yeh, I think we can do things a lot smaller on the mobile than we realise in general

joeycor92 (approved):
Think it looks class. As mentioned, we will need to update the timestamp to a nicer format. Good call from Becks to make smaller. Happy with this with the condition that the timestamp is changed at a later date.

PR #335, opened 2022-10-13
313 update routes historical data

joeycor92:
Hi Folks, 

As mentioned on the call. This is a stop gap for now. Will update to look nicer over the weekend. For now we have two routes which return the data in the format that Becks needs. 

image

This should be enough to give us our data point for month and year tomorrow.

Thanks


WarrenK-design (approved):
Happy if becks needs it in this format, be yes do agree needs to be changed in future.

beckskelly (approved):
looks good! understand this is a stopgap, id rather approve this now its getting late

PR #336, opened 2022-10-13
Stockprices api connected

beckskelly:
Pulled in data from month api and year. No date bar at the moment until we can fix it


src/components/stockVisualisationComponents/ChartTypes/PriceChart/PriceChart.js
```diff
@@ -10,45 +10,85 @@ import {
...
-        { name: '2017', price: 400 },
-        { name: '2018', price: 10 },
-        { name: '2019', price: 60 },
-        { name: '2020', price: 700 },
-        { name: '2021', price: 100 }]
+        { date: '2022-01-10', price: 400 },
+        { date: '2022-01-11', price: 700 },
+        { date: '2022-01-12', price: 60 },
+        { date: '2022-01-13', price: 700 },
+        { date: '2022-01-1', price: 500 }]
```
BearachB, line 75: Should these be hardcoded?

BearachB (commented):
Are the day values supposed to be hardcoded in?

joeycor92 (changes requested):
Getting a few errors which should be easy to fix.
image

The charts really do look fantastic and the transitions are brilliant when going from period to another. Unreal

src/components/stockVisualisationComponents/ChartTypes/PriceChart/PriceChart.js
joeycor92: They should be, we don't have daily data yet. My weekend job.

beckskelly:
@joeycor92 sorted those errors now! get to da chopper

src/components/stockVisualisationComponents/ChartTypes/PriceChart/PriceChart.js
BearachB: Grand so! Wasn't positive

joeycor92 (approved):
Looks good now. Thanks for getting rid of those errors.

BearachB:
Looks really good, well done @beckskelly!

PR #338, opened 2022-10-15
Reuters article description and 13 new RSS sources

WarrenK-design:
Noticed a problem with the Reuters articles where the description was being stored using the HTML tags in the database. They used two descriptions in 2 <p> tags, the first was always half a sentence, the function getReutersDescription now parses the second <p> and get a proper article description. Added 13 new RSS feeds to parse, the total number of RSS feeds parsed for articles is now 32.


DaraghK93 (approved):
Looks good. Seems like the best way to do it

joeycor92 (approved):
Looks good, as per our discussion we might need to keep an eye on this 
image

As you said, they seem to be coming in ok now but just something to watch for if it was an issue.

PR #351, opened 2022-10-17
Infobutton components created #350

beckskelly:
Created 2x info button components. The first is a standard button with a modal, shown below. The second is one with a hover info component. With the second button ,you can set the placement to have it top, bottom, left or right as a prop.

Also deleted a few comments here and there.

For pushing to the app, I'm using the modal button on the company information section of the stock information page.

image

Modal:
image

Examples of the hover button in action:
Hover with right prop
image

Hover with bottom prop 
image

### DIFF
--- a/src/components/stockVisualisationComponents/ChartCard/ChartCard.js
+++ b/src/components/stockVisualisationComponents/ChartCard/ChartCard.js
@@ -3,9 +3,6 @@ import { useState, useEffect } from "react";
 import BarChartViz from "../ChartTypes/BarChartViz/BarChartViz";
 import PieChartViz from "../ChartTypes/PieChartViz/PieChartViz";
 import RadarChartViz from "../ChartTypes/RadarChartViz/RadarChartViz";
-import { Info } from "react-feather"
-//TODO
-// Move dropdown to right hand side without losing functionality of button
 
 const ChartCard = ({ title, data }) => {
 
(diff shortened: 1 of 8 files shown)


WarrenK-design (changes requested):
Maybe a little bit of a margin on left looks a little squished. 
image

Can we stop using on big css file for component specific css, think this should be broken out ```InfoButtonModal.scss``` file to avoid that ```custom.scss``` getting too big.
image


Things like this ```size``` could maybe be specified in the css class then as height and width, be easier to change later. 
image

Can we change this to something relating to us maybe. 
image

beckskelly:
These are all good spots @WarrenK-design , I resolved the size issue by setting a width and height attribute in the scss file. Adjusted the padding so it looks like this now

image

WarrenK-design (approved):
Looks Good!

joeycor92 (commented):
Sorry didn't get a chance to look before now. Just wondering if these are now fixed for desktop and mobile, and if we should be doing that? Does it affect how responsive our page is?
image

WarrenK-design:
> Sorry didn't get a chance to look before now. Just wondering if these are now fixed for desktop and mobile, and if we should be doing that? Does it affect how responsive our page is? image

That's actually good one, maybe use ```rem``` could see how it works

beckskelly:
@joeycor92 @WarrenK-design didnt think of that, changed to rem there, Im going to set up the scss global variables after this and ill tidy up other units while im at that

image

joeycor92 (approved):
Yeah rem works a bit better. Thanks very much for that change. Happy out now.

PR #354, opened 2022-10-17
Scss global variables

beckskelly:
Moved all the variables into their own file, changed all static values (px) to rem

### DIFF
--- a/src/scss/README.md
+++ b/src/scss/README.md
@@ -1,0 +1,12 @@
+### README:
+
+As variables have a scope, if a variable is only used in one place, 
+they should be placed in their own separate file as below:
+
+scss/
+      globalVariables.scss
+      /sampleComponent/
+                     sampleComponent.scss
+                     sampleComponentVariables.scss 
+
+Otherwise, place them in the glbal_variables.scss file and state where they are used in the app.
--- a/src/scss/custom.scss
+++ b/src/scss/custom.scss
@@ -13,13 +13,8 @@
 
 @import "./widgetsComponents/infoButton.scss";
 
-
-
-$primary: #006195;
-$background: #ffffff;
-$text-colour: #020202;
-$submit: #e16202;
-
+/* Variables */
+@import "./variables.scss";
 
 body {
     background-color: $background;
@@ -56,14 +51,14 @@ h1, .h1 {
 }
 
 .stockInfoCol {
-    margin-bottom: 10px;
+    margin-bottom: 0.625rem;
 }
 
 .infoCardStyle {
-    padding: 10px;
-    margin: 10px;
-    margin-bottom:20px;
-    padding-bottom:50px;
+    padding: 0.625rem;
+    margin: 0.625rem;
+    margin-bottom: 1.25rem;
+    padding-bottom: 3.125rem;
     display:flex;
 }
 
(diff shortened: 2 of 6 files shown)


WarrenK-design (changes requested):
I like this but can we break it up with comments somehow. or something by end of project can see this file being quite big 
```
/* File Name One Variables Used */
varOne= 
VarTwo= 
/* File Name Two Variables Used */
myVar=
nextVar=
/* File Name Three Variables Used */
somethingElse=
```
Or an alternative is to have 1 global variables file, this holds styles which are used throughout the project, things like primary, secondary colour. 
If a variable is then used only for a specific component then it should be in its own variables file or just defined within that file. Something like this below or just have definition of local vars in the scss file they are used maybe? 
```
scss/
      globalVariables.scss
      /sentimentComponents/
                     sentimentBadge.scss
                     sentimentBadgeVariables.scss 
```

joeycor92 (approved):
All much better with REMs instead of fixed values. I can see what Warren means, it's good to have as many comments as possible throughout just especially for us reviewing. Great stuff.

beckskelly:
@WarrenK-design I'll defo add in comments as to where all these variables have gone, it will aid us with readability down the line.

For variables in a specific component, I think it's better to have them all in just the one global variables file, as from a point of view of design, a reduced colour palette is best. So for example, your positive, neutral and negative colours for your badges would be resued again in the graphs.

I am reading around extracting the variables at the moment so they can be used in the graphing package easily, as currently they can't without just hardcoding into the code. Because of this, I don't think we will have more than 20 colours max, (if we do shoot us and our awful rainbow app) and I think they should stay in the one global_variables file.

WarrenK-design:
> @WarrenK-design I'll defo add in comments as to where all these variables have gone, it will aid us with readability down the line.
> 
> For variables in a specific component, I think it's better to have them all in just the one global variables file, as from a point of view of design, a reduced colour palette is best. So for example, your positive, neutral and negative colours for your badges would be resued again in the graphs.
> 
> I am reading around extracting the variables at the moment so they can be used in the graphing package easily, as currently they can't without just hardcoding into the code. Because of this, I don't think we will have more than 20 colours max, (if we do shoot us and our awful rainbow app) and I think they should stay in the one global_variables file.

Ah yeah for colours I think this is fine to have things like "positive" "negative" "neutral" in the global but just thinking down the line from reading into scss, variables can be [more than colours](https://sass-lang.com/documentation/values) and they do have [scope ](https://sass-lang.com/documentation/variables#scope). I don't think defining everything in global scope if we really get into using it is best approach, if a variable is only used in one place should be scoped to that file and not in the global scope.

beckskelly:
@WarrenK-design  Ok cool! All good points especially on variables not just being colours and their scope. As I intend on using the colours you defined in your news sentiment scss file (and so will Caolan) I am moving them to the global_variables file, but I will also include a readme comment at the top explaining that if a variable is used in one place only to use the structure as suggested in your initial comment.

beckskelly:
@WarrenK-design is this explanatory enough? 
image

joeycor92 (approved):
I would think that's probably enough detail that you have their now so I'm happy with it.

WarrenK-design (changes requested):
Yep nearly perfect but if doing readme do it in README.md format not in comments, so in file ```src/scss/README.md``` bit handier to update down the line than comments.

beckskelly:
@WarrenK-design Grand! added the README.md

WarrenK-design (approved):
Looks good!

PR #356, opened 2022-10-17
Add daily data function to Lambda

joeycor92:
Hi Folks,

Created the lambda function. This isn't the easiest code to read through so let me know if ye are having issues. 

It's supposed to run every 20 minutes using cron.

It takes about 80 seconds to run so you'll need to use 
amplify mock function --timeout 80

You can see that in the database test.sample_stock_data that the two fields are being updated.

Price field
image

Daily_change
image

I'll update to send to the actual database when reviewed.


WarrenK-design (changes requested):
stockPriceFunction is the only one with code in it, can you get rid of anything you dont need. 
image

Duplicate code. 
image
image

Need to be able to run this locally against a local db not this Mongo URI. You have copied README which has environment variables to do this but haven't implemented the "dev" or "prod" check.  
image
image

Need comments for your functions, use [docstrngs ](https://peps.python.org/pep-0257/) too just makes things more maintainable and used in nearly all Python packages, this is vs code extension which generates them for you. 
image

Definitely need comments for edge cases like this why do we do this if "i" is "BRK.B" and why is a KeyError acceptable? 
image

Haven't ran it yet but these are some of things that need to be fixed, dont really want to run it either if it will only go to production database.

joeycor92:
> stockPriceFunction is the only one with code in it, can you get rid of anything you dont need. image
> 
> Duplicate code. image image
> 
> Need to be able to run this locally against a local db not this Mongo URI. You have copied README which has environment variables to do this but haven't implemented the "dev" or "prod" check. image image
> 
> Need comments for your functions, use [docstrngs ](https://peps.python.org/pep-0257/) too just makes things more maintainable and used in nearly all Python packages, this is vs code extension which generates them for you. image
> 
> Definitely need comments for edge cases like this why do we do this if "i" is "BRK.B" and why is a KeyError acceptable? image
> 
> Haven't ran it yet but these are some of things that need to be fixed, dont really want to run it either if it will only go to production database.

Thanks for all these, I think I've fixed them all. For the keyerror, I haven't got one yet, but I do think that if the S&P500 changes, i.e. adds or deletes funds, we might have an issue. 

Something for us to think about.

WarrenK-design (approved):
Looks good! Just when do merge remember to try run it on AWS Lambda and check cloud logs

PR #361, opened 2022-10-18
Backend - stock discovery categories - backend controllers for Top ESG companies

DaraghK93:
I've separated out the logic for sorting top E, S and G companies into stockServices. I think this is the best way to do it.

Stock Controller
Screenshot 2022-10-18 at 12 09 32

StockServices 
Screenshot 2022-10-18 at 12 10 08

I'll quickly do this for the rest of the categories today. just want to make sure I'm doing it in a way that won't slow down the loading

As per a chat with Caolan, the getAllStocks controller will return an array of sub arrays, each containing the stocks that will be displayed in each separate React side scrolling menu components.


WarrenK-design (changes requested):
The using of services is good but I think you can get away with 1 query and deleting all the services with using a few functions mongo has, the database will always be quicker at doing things like sorting/filtering/searching and we are already struggling with loading times of stocks. 

The [$group aggregator](https://www.mongodb.com/docs/manual/reference/operator/aggregation/group/). This will return groups of stocks instead of you querying all the stocks. 

There is also a [$sort aggregator](https://www.mongodb.com/docs/manual/reference/operator/aggregation/sort/) which would be useful in sorting groups. 

Then [$limit](https://www.mongodb.com/docs/manual/reference/method/cursor.limit/) function too so you aren't getting all stocks because at moment you are querying db for all stocks. 

One database query where the database performs what the services are doing would be mre optomised. 

$Group by E>Value, S>Value, G>value, $sort descending, $limit 20. 

I tried to run a GET request too and getting this in postman
image

caolan-power (commented):
Yeah I'll echo Warren. As much that can be done mongo side the better. Remember we looked at those queries the last day when we thought you'd have to return the whole 'prices' object and it was too slow. That could have an answer in there

DaraghK93:
Thanks, I'm just researching how to do this now. so this should be one query in the stock controller that returns 3 arrays of documents in the relevant order?

caolan-power:
The way I imagine it would be similar to ORDER BY ascending or descending in SQL by a certain parameter (E,S,G) in your case. Then you can apply .limit to return only what you need. I'm not exactly sure what is possible to be honest but even if you had to make 3 separate calls to the db for 20 stocks each it's way better than 1 call for 500. 

    const stocks = await Stock.find()
      .select({ prices: 0 })
      .sort({ 'esgrating.environment_score': -1 })
      .limit(20);

Something like this would work. If you can put all into one query and aggregate results even better. But even 3 calls of something like the above would work for now. Way quicker

DaraghK93:
I've made it all just one Mongo query now as below
Screenshot 2022-10-18 at 20 11 11

Let me know if you think this would work. I can also get all arrays of stocks needed for the discovery page in one query if needed. might be the best way to do it. it's still a little slow on postman but faster than  through nodeJS

WarrenK-design (changes requested):
Hi Daragh the query looks really good and have run it on database seems to be working there just few things going on when I run the website now on the stock discovery page.

image

Can see the image one anyway think you are excluding that in the query, the console error you just need a key for them news sentiment cards I think. Not sure why only 3 articles being returned though on screen with no info about companies. 
image

For this one too if you wanted you already set the stock state in line previous, if you didnt want to mange more state could directly access from the stock state ```stock.esgrating.environment_score``` for example wherever you need it on the page, will save us storing duplicate data. 
image

For this too should there be an ```if``` loop or something here like if I send a particular query param then I get this data? Then if I don't then I get all stocks? Could be if ```summary=True``` in query param or something like that 
image

DaraghK93:
Thanks for the comments Warren. 

I've changed the query to return logo.

I've also changed the getAll stocks function to check the query params and see if category = summary, in which case it returns the different categories of stocks (right now React is just showing the top environmental stocks but this will change). otherwise, it will return all stocks as we had before. This will change when Bearach has implemented searching

Screenshot 2022-10-19 at 15 46 20

so our app.js now brings us to the stockdiscovery screen for both the above paths, however within the screen, a conditional will check the paths to decide on which stocks to show.

Screenshot 2022-10-19 at 15 47 39

WarrenK-design (approved):
Looks a good start Daragh, working in postman and on react page. 

Just remove this console log before merging
image

BearachB (approved):
Looks good, can't see any errors/bugs/problems when I run it. Good stuff

PR #366, opened 2022-10-18
Fixing twitter scraping to reflect new displaying requirements

caolan-power:
No issue to attach to this as it's a bit of an impromptu one. 

Need to change the lambda function to scrape extra data from twitter now that I am making the tweet component from scratch. Not a huge PR


caolan-power:
Forgot to turn formatter off editing files that weren't mine. Only real changes are to tweet model, tweet controller and index.py in twitter scraping lambda. sorry

amplify/backend/function/expressServer/src/controllers/stockController.js
```diff
@@ -24,7 +24,7 @@ const getStockPrice = async (req, res, next) => {
   }
 };
 
-// @route   GET api/stocks
+// @route   GET api/stock
 // @desc    Get all stocks
 // @access  Private - add auth middleware to make it private
 const getAllStocks = async (req, res, next) => {
```
BearachB, line 30: @caolan-power Is this getAllStocks function supposed to be here?

caolan-power: Nope and can't see it on my local machine so strange

caolan-power: should be ok now

BearachB (approved):
Looks good overall, recreating the Tweets will give us a lot more flexibility in what we display!

PR #367, opened 2022-10-18
Added password length checker to registration screen with minimum len…

BearachB:
…gth enforcement of 8 characters. 

# Running with no errors

running_no_errors

# Changes

- Added password length check with red error popup
password_length_error_msg

---
# Future Work

- I tried to add in a password length checker to the password field, most of this work was done however I ran out of time to get it fully implemented and need to move on. I have these changes saved locally and will eventually add this change but for now I need to move onto the stock discovery search feature. 
- When I am doing this I will also update the registration form to use REACT components, as it stands it's quite hard to adjust and update elements because all the fields are hardcoded into the registration page.

It will look something like this:
image

### DIFF
--- a/src/screens/registration/registrationScreen.js
+++ b/src/screens/registration/registrationScreen.js
@@ -35,7 +35,7 @@ function RegistrationPage() {
   const [username, setUsername] = useState('')
   const [password, setPassword] = useState('')
   const [confirmPassword, setConfirmPassword] = useState('')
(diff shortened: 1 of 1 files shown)


DaraghK93 (approved):
looks good, let me know when the future changes have been implemented

beckskelly (approved):
Looks grand, hope to see the strength checker implemented some time when you have time!

PR #368, opened 2022-10-18
Stockinfolayout

beckskelly:
Updated layout of stock individual page basic information at the top of the page. Got rid of background and changed font and logo size. Let me know what you think.

Phone:
image

Desktop:
image


BearachB (changes requested):
Looks really good Becks! Starting to take shape.

One thing I noticed is that pinning the "Trade" button to the bottom covers over the cards at the bottom. See screenshot, Might need to change this so that it's doesn't overlap (more like a footer)?

image

beckskelly:
Hey @BearachB , I see the issue. I cant just have it like a footer under everything because its supposed to be viewable all the time. I'll see about inserting some blank space under the page to try push the other elements up. Good spot

beckskelly:
@BearachB fixed that now 
image
image

BearachB (approved):

joeycor92 (changes requested):
Maybe change the padding or something around this?
image

Also there's a lot of white space still. I just wonder if we moved Symbol, Total Value and Movement to one row. It would reduce the space an awful lot

beckskelly:
sorted that padding there @joeycor92 , that was a good spot, I need to check out different logo's in the pages in the future, theyre all different so it can be difficult!

Regarding the white space I know what you mean, I am intending on adding a follow button later in the day, the intention was to put it on the right hand side (as per prototype), but space might be an issue here. Would you be ok with me expirimenting with the layout of the information when im inserting the follow button, as it will inevitably change again when I get to doing that?


for reference heres the prototype im copying
image


and here are the updated changes
image

joeycor92 (approved):
Completely forgot about the follow button! happy for you to experiment and then just get the opinion of 2-3 of us each time. but you're right in general to follow the prototype. Good to have a mess around with these things.

PR #369, opened 2022-10-18
Fix amplify dependent file

joeycor92:
Updated amplify dependent file which is causing the build to fail.


caolan-power (approved):
All good. Hopefully this fixes it

DaraghK93 (approved):
good stuff. hopefully this works

PR #373, opened 2022-10-18
Stock data intra day prices

joeycor92:
Hi Lads,

This is the exact same as before except a new branch and I don't think any of the previous mistakes have happened again.

Let me know if ye see anything mad.


DaraghK93 (approved):
looks good. hopefully it's sorted this time

amplify/backend/function/stockPriceIntraDay/amplify.state
```diff
@@ -0,0 +1,6 @@
+{
```
caolan-power, line 1: this was the file causing issues last time too so hopefully

caolan-power (approved):
Exact same as last time so accepting

PR #376, opened 2022-10-18
Stock data intra day prices

joeycor92:
Have updated the env variables there. Hopefully all good now.


DaraghK93 (approved):
that should do it I think

caolan-power (approved):
Nice including the environment variables

PR #378, opened 2022-10-19
Stock info set crontime

joeycor92:
This is so that it will run every 20 minutes from Monday to Friday between 9am-10pm Irish Time.

As per the below which shows some examples of the next 10:
image


WarrenK-design (approved):
Looks good

DaraghK93 (approved):
looks good

PR #380, opened 2022-10-19
fixed an error from an usused import 

beckskelly:
I accidently pushed an unused import in my last PR. Sorry about that, got rid of it now

image


joeycor92 (approved):
Grand, strange one though as I had that yesterday and was about to flag but then when I restarted the app it was gone.

BearachB (approved):
Weird, I also didn't see that as an error when I ran it?

PR #381, opened 2022-10-19
Price chart overhaul

beckskelly:
Changed the appearance of the pricechart. The dates look a bit rotten but I decided to not bother changing the format as I believe someone on the backend is going to fix that for me anyway.

Added a tooltip that works on click for mobile, on hover for desktop

I got rid of the "price chart" title because I feel the user will get the picture

I have removed the card background, changed the gradient below the line to a lighter green block colour, have no dates showing on the mobile, I changed the buttons too and the position of them

On the Y axis the prices are showing up after the numbers (as the graph percieves them as "units"), I will change this in the future but would like to move onto the ticker cards for now.

On phone
image

On desktop
image

All runtime errors gone this time as well btw!


BearachB (approved):
Looks much better than before, good job. No errors or bugs noticed.

joeycor92 (approved):
Looks good. The only thing I have, and something I mentioned before is just looking at not starting the graph at 0. Have looked at a few and they don't e.g. 
image

It's only a small thing that we might look at in the future when all the data is coming in.

beckskelly:
@joeycor92 I'll record that in the issues! thanks all for the review

PR #383, opened 2022-10-19
Backend portfolio update portfolio schema

joeycor92:
Update to the Portfolio and User models. The PortfolioSchema is an object within the UserSchema. 

The routes haven't been touched yet but they will be changed next.


WarrenK-design (changes requested):
Quick one so one user can have multiple portfolios so it should be an array. The portfolios array should reference the portfolio object ID. 
```
portfolios : [{
  portfolio:{type: mongoose.Schema,Types.ObjectId, required: true, ref: 'Portfolio'}
]
```

For the actual portfolio object, I persume ```holdings``` is our current stocks, this should be an array which refrences stock objects. 

```
holding: [ 
  balance: {type: float, required: true},
  stock: {type: mongoose.Schema,Types.ObjectId, required: true, ref: 'Stock'}
]
```

Have a read of [this](https://mongoosejs.com/docs/populate.html) for the correct syntax, its basically mongos version of an SQL join/relationships, will make our queries a lot better.

DaraghK93 (commented):
yeah, I agree with Warren. on the holdings, if we're referencing the stock schema it will only need do those calculations once when the price of the stock changes as opposed to putting that logic elsewhere

joeycor92:
> Quick one so one user can have multiple portfolios so it should be an array. The portfolios array should reference the portfolio object ID.
> 
> ```
> portfolios : [{
>   portfolio:{type: mongoose.Schema,Types.ObjectId, required: true, ref: 'Portfolio'}
> ]
> ```
> 
> For the actual portfolio object, I persume `holdings` is our current stocks, this should be an array which refrences stock objects.
> 
> ```
> holding: [ 
>   balance: {type: float, required: true},
>   stock: {type: mongoose.Schema,Types.ObjectId, required: true, ref: 'Stock'}
> ]
> ```
> 
> Have a read of [this](https://mongoosejs.com/docs/populate.html) for the correct syntax, its basically mongos version of an SQL join/relationships, will make our queries a lot better.

This is very handy, I was actually looking for something like this and couldn't find it. I'll have a go at updating this now. Will get to ye asap.

PR #384, opened 2022-10-19
Machine learning news sentiment feature engineering

WarrenK-design:
Feature engineering for the training datasets for news sentiment analysis. 

For this iteration the features used to train the model are:

1. positive polarity score
2. negative polarity score
3. neutral polarity score
4. compound polarity score
5. positive word count 
6. negative word count 

The positive and negative word lexicon are generated using the training datasets, it may be good in future to try existing lexicons for positive and negative words. Next step is training a model on these features, I think these features will be an iterative process based on testing phase but good chunk of reusable functions here.


joeycor92 (approved):
Really well laid out and clear. Googled tokenizing there to understand it but then realised after that you had it fairly well explained. 

The steps are clear and I'm happy to approve this.

caolan-power (approved):
Looks really good and those features are definitely good enough. 

Having the high quality dataset for generating the words list will really help too.

We can maybe work in expanding features with bigrams and trigrams and explore that but can discuss further down the line.

Be interesting to see model accuracy now

PR #385, opened 2022-10-20
Tweet component 374

caolan-power:
PR for tweet component and tweet container. Uses the same container Warren had for news articles jus to keep things the same. 

First real front-end PR so be sure to tell me if there's some structure or something stupid I'm doing to change.

I'm not using CSS because I don't know if I need it but you could suggest things on that front if needed too.


joeycor92 (changes requested):
I think it looks really good, obviously we'll need to come up with a way to filter tweets better as all the PepsiCo tweets are about Pep Guardiola! But looks wise it's a good start.

Just wondering about the white space and if anything can be done about it.
image

Looking at another example and the content starts at the left side of the profile image.
image


Not sure how easily it can be done. 

Also I know I've mentioned the black hearts a few times, but if it's not functional I think it could be removed. Also, I wonder if they clicked on the tweet would it bring them to the tweet on Twitter and they could read further discussions and get involved in heated and pointless twitter debates with random people from the internet.

beckskelly:
also in the future for frontend prs please submit a screenshot with your request in a comment, makes it a lot faster for us for frontend UI

beckskelly:
Just going through your code now, we are not using PX at all where possible, so please convert to REM. Also for your colours you should use the variables set in the global variables. 

image

caolan-power:
Ok cool thanks for all your comments am working on these now

caolan-power:
Screenshot 2022-10-20 at 16 27 20


How are these changes. Heart in or out ? happy with layout?

caolan-power:
twitter bird added 
Screenshot 2022-10-20 at 17 17 13

beckskelly:
I'm happy with it like that! And keep the heart, but I dont mind

beckskelly (approved):
Haappy with it, looks well, thanks for implmenting those changes, I like the little bird in the corner, its very cute!

PR #386, opened 2022-10-20
Stock discovery page search/filter function

BearachB:
# Additions

- Search function added
- This takes input from the user and searches their term across five fields in the DB
  - Symbol
  - Longname
  - Shortname
  - Industry
  - Sector

# Future work

- Sort by - Currently this sorts the results by ticker symbol, however in future the user will be able to chose what to sort by. For this to happen the DB needs to be updated (most fields in as strings and can't be sorted as numbers yet, this is being worked on).
- Filter by - User will be able to filter based on document fields, such as only show results for companies in "airline" industry, or only show companies with less than 5,000 employees etc.

---
# Screenshots
image

## Showing the default discovery view that @DaraghK93 is working on:
image

## Search results for "micro"
image

## Search results for "apple"
image


DaraghK93 (commented):
looks good Bearach.

There seems to be issues when doing multiple searches, like if you search for Apple, then from that page search for something else, it didn't show anything for me. but it worked once.

also, when i search with nothing in the search bar it takes me to a blank page
Screenshot 2022-10-20 at 17 37 43


you might let me know if these are happening for you?

BearachB:
> There seems to be issues when doing multiple searches, like if you search for Apple, then from that page search for something else, it didn't show anything for me. but it worked once.

Hmm ok that is odd, that works for me. Can't replicate the problem. Might try get @joeycor92 to see if he can replicate the issue?
 
> also, when i search with nothing in the search bar it takes me to a blank page Screenshot 2022-10-20 at 17 37 43

Yep that has not been addressed yet, when the user leaves a blank search it will eventually show all. This will be fixed tonight/tomorrow.

DaraghK93 (approved):
I've had another look there and I think it may just need some query optimisation. It works though and looks great!

As well, as we were saying, it would be good to get a keyword list in for each company to help with the searching.

BearachB:
> I've had another look there and I think it may just need some query optimisation. It works though and looks great!

Definitely agree, it's a long way off being finished (filtering and sorting needs to be added) however I think for now, it functions as a basic search function.

> As well, as we were saying, it would be good to get a keyword list in for each company to help with the searching.

Yep definitely need to look into this, a big oversight when a search of "Google" doesn't bring up the company that everyone knows as Google.

PR #389, opened 2022-10-20
follow button and Price chart color

beckskelly:
Implemented the follow button, changed price chart buttons again. Button is now highlighted when selecfted and has a small border around it
Subtle things to the price chart like making it red when the stock is down, green when up. Havent changed it for entire years worth would like to get those ticker cards out today instead.

Made the stroke line fatter on the pricechart which highlights it more

Let me know thoughts

image

image

image


BearachB (approved):
Looks really good! Taking shape slowly but surely.

Just on the visual side of things, not saying to add this in now but maybe keep in mind for when you come back to this in future:
- 1D, 1W, 1M, 1Y buttons - Could make the colour of the border to match the colour of the Y axis ticks/values (the price in USD).  
- Same as above for the X axis ticks/values

BearachB:
Just to confirm as well before I forget, the dates are going to be fixed eventually to show something neater than what's shown currently?:
image

beckskelly:
Fixed that, you've a good eye for detail so thanks for that spot. Yes, the dates will be neatened up in a route later on so I've left them like this

DaraghK93 (approved):
Looks great becks,

Just adding to Bearach's, I agree on the dates at the bottom, these will need to be changed.

Just the prices on y axis, can we make them '$20', as opposed to '20$'?

beckskelly:
Hey @DaraghK93 , Im gonna look into the prices later, the package registers them as units so its being a bit annoying, but its one I want to resolve in the future!

PR #391, opened 2022-10-20
New tickercards

beckskelly:
Redid the ticker cards based on Bearachs prototype
Basically a big restructure of the info, I included in some of our widgets like the sentiment badge, twitter logo, info button
Sorted all errors

Some dummy data like the overall sentiment, should be obvious as its repeated across all cards

Let me know any comments

image

image

image

### DIFF
--- a/src/components/stockDiscoveryComponents/tickercard/Tickercard.js
+++ b/src/components/stockDiscoveryComponents/tickercard/Tickercard.js
@@ -2,7 +2,7 @@
 //  This component is used to display stocks to the user
 //  It is a card which shows an overview of the stock showing some key details 
 
-import { Card, ListGroupItem, ListGroup} from 'react-bootstrap';
+import { Card, ListGroupItem, ListGroup, Row, Col } from 'react-bootstrap';
 
 import { Link } from "react-router-dom";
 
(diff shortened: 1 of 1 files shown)


BearachB (approved):
This looks really good @beckskelly! (Although I suppose I am a bit biased...)

Think it's an improvement, however one thing I'd say is, I wonder if there is a neater way to show the current price and 24hour price? Again, not something that needs to be changed on this PR, I'm happy with how it looks now but down the line I wonder if we could figure out a cleaner/more intuitive way of showing it. Might be something that we mock up and do some A | B evaluating with people to see what they prefer?

Just for some reference (might start a bit of discussion on it), here is how a few websties show show the same info:
## DeGiro.com 
image

## Yahoo! Finance
image

## Google Finance
image

## Interactive Brokers (please excuse the woeful quality)
image

Seems like the standard way of showing it is just showing the three numbers, however this might not be ideal for us as we're aiming for people who don't have much experience. I wonder if we could instead maybe do something like just showing the three numbers with maybe an information i, that when you click on it gives you a quick explanation of how the 24h prices are worked out.

DaraghK93:
I think they look great Becks. I think the information is very clear, but like Bearach said, some A| B testing could be very beneficial to have the information laid out as clearly as possible.

One thing is that "Did you know section". it's a good idea but we have to think how we're actually going to input that information. the rankings will change and it could be hard to find information we can put on every card

DaraghK93 (approved):

PR #393, opened 2022-10-20
Frontend UI - stock discovery - create horizontal scrolling menus

DaraghK93:
created the stock container component for stock discovery. This will break the stock discovery page up into various rows and display them as horizontally scrolling menus, like the netflix homescreen.

Screenshot 2022-10-20 at 21 06 36

I had to slightly change the styling of the ticker cards as well, done inline on the TickerCard.js file.


amplify/backend/function/expressServer/src/config/default.json
```diff
@@ -1,5 +1,4 @@
 {
 "jwtSecret": "itsasecret",
-"MONGO_URI": "mongodb://admin:pass@ec2-3-249-145-103.eu-west-1.compute.amazonaws.com:27017"
-}
```
BearachB, line 4: Not quite sure why this is showing up as a change, as this new URI is already merged to master. Weird but at least it's not changing

BearachB (approved):
Looks good overall Daragh! I've left some minor comments, nothing big enough to hold off the PR though.

One more minor comment - Might want to add a bit of spacing between the search bar and the title of "Companies That Are Good For The Environment" (see screenshot). Also, we need to come up with some catchier titles for those categories, not sure what they'd be but need to have a think about it.

image

WarrenK-design (approved):
Hi Daragh,

Looks like its coming together nicely, still needs a bit of work but accepting so we can show it tomorrow. Adding top movers, top losers, etc. and fixing the cards a bit more but think Becks nearly done with some of that. 

I did commit to it to just get rid of that annoying key warning so do a pull if you change anything before merging. 

image
image

PR #394, opened 2022-10-20
Machine learning news sentiment model training

WarrenK-design:
Model training started, Python module similar to last few pull requests with reusable functions, within this pull request they relate to training of models. 

3 iterations of models trained using the different features, see the scripts under ```modelGeneration``` to see how these models are being trained. The models are being evaluated only on accuracy at the moment, these reports are written to csv files. The trained models themselves are written to pickle files which will eventually be used on AWS Lambda news scraper function. 

Summary of accuracies so far. 

 
image


joeycor92 (approved):
Bit of reading in that! Great stuff, the steps all make sense and the comments all explain what's happening really well.

PR #396, opened 2022-10-22
Update stock routes

joeycor92:
Right both, have included one backend, one frontend, big PR here. Felt it would have messed up the site if I had done backend and frontend separately but hopefully it's not too bad to check.

Backend:
Updated route api/stock/:symbol - Changed it so that it would return the needed data + the week, month, year data for the graphs. No work needed on the front end to change the variable type from string to float.  The function which does the work is the same as the previous function but is now in a new file in the services folder. Week data is now available as well.

Another change is that the dates now come in the form DD/MM as per below. This will hopefully help with the appearance of the graphs. Can change to DD MMM if needed.
image

Frontend:
No need for two more requests. Removed any requests in the stock price chart as the data is now given in the parameters. Total of one request for the whole page. Also removed the loading spinner as this would stay loading forever without the requests. Removed some use states which were not needed.

image

All seems to be loading fine.
No errors here:
image

nor are there any on the developer tools web console.


beckskelly (changes requested):
Hey @joeycor92 , Looks really well, I mostly reviewed the frontend code as I have little understanding of the backend of the project, but the backend does look logical and efficient. I see it's a big PR but you were better off doing both sides in this instance.

On the frontend its nearly perfect, and the page is loading faster etc - so thanks for that and keeping the site efficient. Dates look so much better too and you got rid of a lot of uneccesary code by just having one call to the backend. 

Just one thing on the frontend, when the page is loaded in desktop mode the dates don't show. They only show when you load in mobile and resize to desktop. 

I had a little bit of messing around just there and got it to work with the following 3 lines and adding the useEffect dependancy back in:

image

So my suggestion is to just add that back in so the dates appear onload in desktop mode. If you can see a better/ more efficient way to do it go ahead, I tried adding a listener to "onload" first like how I did with resize, and it didn't work.

Also I'd delete this comment if we arent using a loading spinner anymore in this component
image



Once again thanks for sorting this out, really good work.

beckskelly:
Oh one more thing before I forget, I'd connect this to an issue for our burn down charts

beckskelly (approved):
Thanks for that change Joey looks all good on my end anyway!

joeycor92:
> Hey @joeycor92 , Looks really well, I mostly reviewed the frontend code as I have little understanding of the backend of the project, but the backend does look logical and efficient. I see it's a big PR but you were better off doing both sides in this instance.
> 
> On the frontend its nearly perfect, and the page is loading faster etc - so thanks for that and keeping the site efficient. Dates look so much better too and you got rid of a lot of uneccesary code by just having one call to the backend.
> 
> Just one thing on the frontend, when the page is loaded in desktop mode the dates don't show. They only show when you load in mobile and resize to desktop.
> 
> I had a little bit of messing around just there and got it to work with the following 3 lines and adding the useEffect dependancy back in:
> 
> image
> 
> So my suggestion is to just add that back in so the dates appear onload in desktop mode. If you can see a better/ more efficient way to do it go ahead, I tried adding a listener to "onload" first like how I did with resize, and it didn't work.
> 
> Also I'd delete this comment if we arent using a loading spinner anymore in this component image
> 
> Once again thanks for sorting this out, really good work.

This has been updated now.

DaraghK93 (changes requested):
Hi Joey, great work there. Page loads so quickly now. Stock routes is nice and tidy as well since you made the services file. I think that's a good way to structure it anyway.

There was a missing import statement in PriceChart.js. needs to import useEffect from React as well like this
Screenshot 2022-10-24 at 11 28 24

so just requesting that change

joeycor92:
> Hi Joey, great work there. Page loads so quickly now. Stock routes is nice and tidy as well since you made the services file. I think that's a good way to structure it anyway.
> 
> There was a missing import statement in PriceChart.js. needs to import useEffect from React as well like this Screenshot 2022-10-24 at 11 28 24
> 
> so just requesting that change

Thanks, updated now.

DaraghK93 (approved):
looks good, thanks Joey

PR #398, opened 2022-10-22
Stock price data in number format

joeycor92:
Very small change to the lambda function. Just a matter of processing the data to come in number rather than string. For percentages, Daragh noted it would be easier for him to sort top movers if the data comes as float. 

When used on the front end a '%' will need to be added.

Mock function looks like this (positive and negative number example):
image
image


WarrenK-design (approved):
Looks good, better sorting it here than later after it has been stored in DB

DaraghK93 (approved):
Perfect, thanks Joey, will make it a lot easier to sort

PR #402, opened 2022-10-24
fixed ribbon and ticker card css

beckskelly:
Fixed ribbon and ticker card css so they cohabit nicely now. Aware its not completely by the prototype but this is the best I can do for now. I made the images a lot smaller as well as they just wont play ball, we need to try resize these at some point

image

image

all errors resolved

### DIFF
--- a/src/components/stockDiscoveryComponents/tickercard/Tickercard.js
+++ b/src/components/stockDiscoveryComponents/tickercard/Tickercard.js
@@ -42,7 +42,7 @@ function TickerCard({ stock }) {
               src={stock.logo}
               style={{
                 display: "block",
-                width: "60%",
+                width: "65%",
                 margin:"auto"
               }}
             />
--- a/src/components/stockDiscoveryComponents/tickercard/Tickercard.js
+++ b/src/components/stockDiscoveryComponents/tickercard/Tickercard.js
@@ -43,6 +43,7 @@ function TickerCard({ stock }) {
               style={{
                 display: "block",
                 width: "60%",
+                margin:"auto"
               }}
             />
           </Card.Body>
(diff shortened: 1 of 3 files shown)


BearachB:
I'm getting compilation warning for unused variables that are defined in Tickercard.js. Could you remove them please.

image

---

Do we think it would be worth horizontally centering the company logos on the card, like below? I think it looks much neater.

image
image

---
Not sure if it's worth it, but might be worth creating a custom scrollbar for desktop at some point down the line, the standard Chrome one looks a bit out of place on the slider ribbon (maybe that's just me though).
image

---
Looks good overall though! Definitely need to revisit the images.

DaraghK93 (changes requested):
Hey Becks, they look so much better than they did, thanks for fixing.

Just one main thing. could you change these on the frontend?
Screenshot 2022-10-24 at 14 42 29

There should be a colon after "change (24h)" and a "%" sign after the absolute change.

Screenshot 2022-10-24 at 14 44 12

also, not a big issue, but getting this warning about row and col. if you don't think you'll be using them, get rid.

beckskelly:
@BearachB @DaraghK93 , thanks for the comments, will fix these now. Oddly enough I'm not getting those warnings in my own browser, mightnt have pushed those changes, ill give it a bash now

beckskelly:
Centered those images and added that % symbol and : @DaraghK93 @BearachB 

image

Still getting no warnings after restarting the app. Could ye both check on your end? I did remove them as part of the pr, see below, so its odd its coming up for you guys

image


Also I agree on the side scroller being a bit windows 1995. Defo something to look into down the line, I'll add it to my list

BearachB (approved):
Good stuff, looks great! Just need to figure out how to handle the images in future.

No errors on my end anymore!

PR #405, opened 2022-10-24
Stock price data in number format

joeycor92:
Format of the date is coming in with seconds so needed to be updated to remove the seconds. 

Daragh also noted that the data for BF-B had the same issue as BRK-B.


caolan-power (approved):
Looks good. Didn't pull this down locally as seems like a small change.

DaraghK93 (approved):
looks good

PR #409, opened 2022-10-25
Added password validation on backend, tested with Postman

BearachB:
# Changes

- Added password validation on the Express side. Now, strong passwords are being enforced on both frontend and backend.
- Exact same validation as on front end, min 8 chars, upper lower and number all needed. 
- Error message is condensed into 1 message on backend, whereas on frontend the user will be told which of the 3 terms they are not meeting.

# Postman Screenshot - Incorrect password format
image

# Screenshot of React terminal
image

### DIFF
--- a/amplify/backend/function/expressServer/src/controllers/userController.js
+++ b/amplify/backend/function/expressServer/src/controllers/userController.js
@@ -35,11 +35,7 @@ const registerUser = async (req, res, next) => {
           'The client has not sent the required information to register the user',
         ),
       )
-    } else if (typeof overeighteen == false) {
-      res.status(400)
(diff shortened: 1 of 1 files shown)


DaraghK93 (approved):
Looks great, thanks. Think that one if statement for password works too and tidies the code up nicely.

BearachB:
> Looks great, thanks. Think that one if statement for password works too and tidies the code up nicely.

Very true, it was actually your comment when I was on a call with you that made me rewrite it that way. Thanks for the input!

beckskelly:
Looks good Bearach! sorry for the slow approval

beckskelly (approved):

PR #412, opened 2022-10-25
Changed port from 3000 to 3001

WarrenK-design:
To try and combat the error of the port being in use already. Also for local dev React uses port 3000 this will fix this issue.

image

 Everyone needs to update there aws-exports.js file to this once its approved . 

image


DaraghK93 (approved):
thanks for this Warren, hopefully will fix the error, but will help with React Port at the very least

PR #413, opened 2022-10-25
Order type component

beckskelly:
Added 4 non functional buttons to the page, also a slider, on click they show blue (see screenshot) but they dont stay that way as the values are currently not held anywhere., down the line they will stay clicked. Added the range slider, this takes in a dummy portfolio value, but does calculate what happens to the portfolio value when you push up and down the slider changing the quantity. Also sorted a gutter issue in the stock discovery page, Warren noticed it just there and its a quick fix so i threw it in

Ignore the desktop for now it looks terrible, I'll fix the desktop when I have all the components on the page

image
image


src/screens/orderConfirmation/orderConfirmation.js
```diff
@@ -25,25 +28,25 @@ function OrderConfirmationPage() {
...
                 const res = await API.get(APIName, path)
                 // Set the state for the stock and loading to false 
-                setStock(res[0]);
-                setLoading(false);
+                setStock(res)
+                setLoading(false)
             } catch (error) {
                 // Log the error 
-                console.log(error);
+                console.log(error)
```
BearachB, line 39: Might want to remove this console.log

src/screens/orderConfirmation/orderConfirmation.js
```diff
@@ -25,25 +28,25 @@ function OrderConfirmationPage() {
...
-                setError(error.response.data.errormessage);
-                setLoading(false);
+                setError(error.response.data.errormessage)
+                setLoading(false)
             }
         }
         getStockInfo();
     }, [])
 
+    console.log(stock);
```
BearachB, line 48: This too

BearachB (changes requested):
Just 2 comments on console logs, not a big deal but could you remove

beckskelly:
Done! Nice spot @BearachB , thanks for that. @joeycor92 any chance of a review today?

joeycor92 (changes requested):
One thing we should fix is that this should really come out as two decimal places:
image

I also think for quantity (you may have this planned later on down the line) but the user should also be allowed to put in the exact amount they want to purchase, e.g. I want to buy 60$ worth of apple, so that's maybe 0.75 shares or something.

Also maybe just a little bit of padding above these 
image

Other than that, really good start on it.

WarrenK-design:
Just a quick thing maybe on this too, from the user evaluation, they taught initially that these buttons up the top would be to actually buy something. I know we have sticky button now too but wondering if something more clear might be needed what you guys think? So stands out as option more or something 
image

WarrenK-design:
$ in front of amount too maybe? 
image

WarrenK-design:
I wonder if "Shares" would be better than "Quantity" too not sure.

beckskelly:
@WarrenK-design thanks for the comments! nice spot on the $ sign, I know what you mean about those buttons, its why I styled them the same as limit order and market order, I thought with the styling they would look more like options (i.e. theyre not filled in.)

Thinking about it, Maybe I should move them into the same component as order type, like this:

image


If you can imagine theyd go blue once the values are held/ assigned, right now they do nothing. So youd have a limit order button selected and a buy one selected, for example.

beckskelly:
Ignore the styling under those buttons I'm trying to get the form to work changing the slider atm

beckskelly:
sorted the console logs @BearachB and thanks to @WarrenK-design and @joeycor92 for the suggestions in changing the design. I now have an input box as well for the user to input an exact amount, the slider changes with the input, and vice versa (the input changes on changing the slider) and does error handling as well

image


image


image

beckskelly:
I see I have conflicts with master I'll sort this now

joeycor92:
> One thing we should fix is that this should really come out as two decimal places: image
> 
> I also think for quantity (you may have this planned later on down the line) but the user should also be allowed to put in the exact amount they want to purchase, e.g. I want to buy 60$ worth of apple, so that's maybe 0.75 shares or something.
> 
> Also maybe just a little bit of padding above these image
> 
> Other than that, really good start on it.


> sorted the console logs @BearachB and thanks to @WarrenK-design and @joeycor92 for the suggestions in changing the design. I now have an input box as well for the user to input an exact amount, the slider changes with the input, and vice versa (the input changes on changing the slider) and does error handling as well
> 
> image
> 
> image
> 
> image

It's actually class, except for one issue that I'm having. 
image
I had a number in there and then deleted and it ended up just stopping on 1.00 and didn't go back to 0.00. I can move the slider to 0 but just wondering why that might be happening.
image
Might be something to with the above code?

beckskelly:
Sorted that @joeycor92 , nice spot so thanks for that

joeycor92 (approved):
Tried a few different things to mess this up and couldn't!! So that's good. Serious work on this.

BearachB:
Looks really good @beckskelly, good job! Slider looks great! If you are looking for more customisability in the slider I tested 4/5 packages on Monday/Tuesday when trying to find one for the ESG ratings, so let me know.

BearachB (approved):

beckskelly:
Thanks @BearachB , I will definitely take your offer on that next week as I've a few ideas for how I would prefer it to look. Merging with main now

PR #419, opened 2022-10-26
Machine learning news sentiment deployment

WarrenK-design:
This PR deals with the machine learning deployment to Lambda, it sets u the flow for deploying models even though this is probably not the final model we will use. 
There are three areas which have been changed. 

# newsArticleScrapingAndSentimentAnalysis
This is the lambda function itself. This had to be updated to first extract the features the classifier was trained on for the real news headlines and then classify them.
## New Files
Within the folder ```src/lib``` directory the words lists used for feature engineering have been added along with the pickle file ```MLPClassifier.pickle``` which is the classifier itself, it was chosen for the first deploy as it received the highest accuracy. 

A new folder ```src/nltk_data``` has been created. NLTK uses its own downloader and locally on your machine will store files under a location like ```C:\Users\<USER>\nltk_data\```. This will not be available on the cloud, this folder just contains the zip files from the local directory that nltk needs, you tell it to look in this directory now in line 32 if ```index.py``` the line ```nltk.data.path.append(f'{dir_path}/nltk_data')```

## Code Changes
The changes to the actual Lambda handler are not really new functions, a lot of them are taken from the ```featureEngineering```` functions developed to train the model. The code just uses these functions to get the features for the live news headlines and then loads the pickle file and classifies them. 

# machineLearning - Local 
Under the machine learning local directory a new file has been created in the resuable functions for sentiment analysis under ```sentimentAnalysis/bin/database```. The functions in her can be used to get a dataabase connection and update the database. 

Under the folder ```sentimentAnalysis/newsSentimentAnalysis/datanbaseUpdate``` a new script ```main.py``` has been created. To run the script read the ```README.md``` in this directory. This script will query the production database for article headlines, classify the sentiment of them and update the database. This script is used to update the database if a new classifier is used. 

# Frontend 
Small chane made to the ```SentimentBadge.js``` component. It now no longer matters if sentiment passed in is lower case or upper case. The check for sentiment will work wither way but the actual display will be uppercase.
image


machineLearning/sentimentAnalysis/bin/database/databaseFunctions.py
```diff
@@ -0,0 +1,55 @@
+import pymongo
+
+
+def getArticles(client,databaseName="test"):
+    """
```
BearachB, line 5: I know you're aware, but just need to make sure we update this (and all other references to the test DB) to the actual DB name when we changeover (likely next week after the interim presentation).

caolan-power (approved):
Hate to give such a small reply to such a big PR. 

It works and I can't see anything jumping out at me to change. The Lambda already worked and the logic of the ML in general worked as per the last PR so that's still ok.

As for setting up the dependencies in virtual environment I will get you to show it to me to save time as there is a lot there you seem to have figured out ( the nltk packages etc ) and how best I can organise my functions etc.

It looks good though and nothing jumping out at me as I say. Great to have it done. Well done

BearachB (approved):
All looks good to me @WarrenK-design. Serious effort on this, well done! Only comment I had was on the DB name, will need to be updated from "test". Probably should have done this earlier, but anyway we can migrate next week.

machineLearning/sentimentAnalysis/bin/database/databaseFunctions.py
WarrenK-design: Yeah can go through next week and update these, only local script the Lambda still working off parameter store so wont be as bad. Cheers!

PR #423, opened 2022-10-26
Updated some image urls relating to updated logos

BearachB:
Minor change to master CSV. No ZenHub issue for this.
- Updated some image urls where the logo image had to be changed (Alphabet, Marsh, Kimberley etc.)
- Also removed some of the weird characters (€ etc.) from the sector and industry columns
- (Updated after @joeycor92's comment on Discord) The urls for the logos in the database were manually updated separately.

This CSV hasn't been needed to update the live database yet, however when we migrate to the final production DB it'll be used to populate it.


DaraghK93 (approved):
great, will look a lot better with proper logos

PR #425, opened 2022-10-26
Portfolio schema, create portfolio route, update user info route for portfolios

joeycor92:
Ok folks. 

New Portfolio Schema, which is nested within the User Schema.

Create Portfolio Route (have tested it and have added portfolios through postman to my own mongo)
image

Updated the getUserInfo route for user so that it returns the portfolios.


amplify/backend/function/expressServer/src/config/default.json
```diff
@@ -1,4 +1,5 @@
 {
 "jwtSecret": "itsasecret",
-"MONGO_URI": "mongodb://admin:pass@ec2-3-249-145-103.eu-west-1.compute.amazonaws.com:27017/"
```
WarrenK-design, line 3: Is this correct or is that your local one? This will change it for everyone.

WarrenK-design (changes requested):

amplify/backend/function/expressServer/src/controllers/portfolioController.js
```diff
@@ -0,0 +1,38 @@
...
+        portfolioName: req.body.portfolioName,
+        remainder: remainder,
+        dateCreated: dateCreated
+      });
+      await User.updateOne({ _id: req.user.id }, {$push: {portfolios: newPortfolio}})
+        
+      res.json({ newPortfolio });
+    } catch (err) {
+      console.error(err.message);
+      res.errormessage = 'Server error';
```
WarrenK-design, line 29: Can we do maybe "Error in creating Portfolio" or something maybe just make it bit easier to debug.

DaraghK93 (changes requested):
Hi Joey, looks great. got a portfolio created in postman there. just take out portfolios from register user, as they don't send any value with it when registering. the model will ensure that's created as default anyway.

Screenshot 2022-10-27 at 10 22 23

Also thinking there should be a check on portfolio name to ensure that two portfolio's related to the one user aren't called the same thing. not urgent for now but might be an idea

joeycor92:
> Hi Joey, looks great. got a portfolio created in postman there. just take out portfolios from register user, as they don't send any value with it when registering. the model will ensure that's created as default anyway.
> 
> Screenshot 2022-10-27 at 10 22 23
> 
> Also thinking there should be a check on portfolio name to ensure that two portfolio's related to the one user aren't called the same thing. not urgent for now but might be an idea

Updated now.

DaraghK93 (approved):
looks good, thanks for changing

PR #426, opened 2022-10-26
Crated a Lambda layer named SentimentLayer

WarrenK-design:
Layer only has 1 package for sklearn at the moment, want to see if we can even just get this up through amplify cli.


DaraghK93 (approved):
looks ok, worth trying

caolan-power (approved):
worth a shot

PR #429, opened 2022-10-27
pushed error, resolving now

beckskelly:
Sorry about that lads ill be more careful next time!

image
image


joeycor92 (approved):
Sorry that I missed this. All good.
Thanks.

BearachB (approved):
Definitely should have spotted that in the review, apologies. Goes to show how easy it is for stuff to slip through reviewing PRs when tired. We're all quite tired at this point, the long days taking their toll, need to be more careful about this going forward.

PR #430, opened 2022-10-27
Revert "Machine learning news sentiment deployment"

WarrenK-design:
Reverts DaraghK93/stockApp#428


BearachB (approved):
Unfortunate you can't deploy it this way, hopefully Docker is the answer. I think the recommender system might be OK to deploy using Lambda because it's going to be a lot simpler of an ML model.

caolan-power (approved):
That's a pity

PR #431, opened 2022-10-27
stock-searching-backend-and-frontend

DaraghK93:
searching now fixed, thanks mainly to Warren.

changed the search results to render results in sides scrolling menus as the ticker cards were being squished together between large screen and phone size screen so this should sort that for now.

Screenshot 2022-10-27 at 16 32 41

Searching for nothing will just bring you to main stock summary page.

Screenshot 2022-10-27 at 16 33 30

All running with no errors. 

Still some styling to be done


WarrenK-design (approved):
Hi Dragh looks good , yeah still bit to do styling but search seems to work. 

For this if no search results use the MessageAlert component.
image

I think the search button too needs to be next to each other.
image

Needs padding too here on mobile @beckskelly not sure if you were working with this. 
image

joeycor92 (approved):
Have conducted as many tests as possible on this and only one thing to write home about. The scroller should probably disappear if nothing is showing up.
image

As discussed, happy to approve this with the intention of sorting that when we can.

Sound.

beckskelly:
@WarrenK-design I defo added padding at one point in a commit but may have undone it down the line, maybe with that mishap I had with git on tues where my commits got pushed but undone. Ill add it back in for the next pr that im doing today, thanks for the spot

PR #433, opened 2022-10-27
trying to commit again. All commits gone because of a git error

caolan-power:
So I had a git malfunction on my end so barely any commits here.

Basically all twitter sentiment analysis and data cleaning, feature engineering done. 

You can run the train_model file to test but the data cleaning won't work because i had to omit the larger dataset since it was too big. I'll go back and re-upload next week.

Just wanted to show the work i've done for the week as this will be my first pull request


caolan-power:
Actually the train model does work so might as well add you both and you can take a look at it. Data cleaning doesn't need to be run for now anyway

WarrenK-design (changes requested):
Hi Caolan, looks good checked it out, can you update the README.md file under the Twitter Sentiment directory. Just explain what each subdirectory is maybe, for running the ```process_data.py``` too what directories do we have to create and what files we have to download can be put in here too,  I know the datasets are too big to store in Github but just maybe give instructions on how we could get this working, I'll check it out again and give it a go.

WarrenK-design (approved):
Looks good

BearachB (approved):
Serious work in here Caolán, well done! Comprehensive readme made it easy to understand the structure. Apologies for not reviewing this for so long.
I'm probably going to need to bug you about the structure/setup of it as I'm working on the recommender system at the minute but it's pretty cobbled together into a Jupyter notebook right now for ease of development.

PR #434, opened 2022-10-27
Frontend UI stock components esg ratings slider

BearachB:
- Quite annoying, I lost the entire description I had written a few minutes ago so I might be leaving some things out here.
Also, not sure who apart from @beckskelly should be reviewing this so I've thrown everyone on. Whoever gets there first I guess.
- Package-lock conflicts because I had to add a new package for the sliders. Also make sure you do an npm i before running locally.
- I also couldn't link the ZenHub issue to this PR for some reason? When I attempt to it just stays loading indefinitely.
- You might need to use the "--legacy-peer-deps" flag to the install the package. 

# Changes
- Replaced the ESG card with a new ESG slider, not quite the Evilometer that was envisioned in the prototype, but it's a step in the right direction.
- These cards consist of 3 components, one slider for each of the ESG ratings.
- Each one is rated out of 5.
## ESG Card Screenshots:
image
image

# Future work:
- Another slider for an overall ESG score should be added.
- Governance svg image is not the clearest, I can update this very quickly once I find something more suitable.
- Ratings need to be normalised between lowest and highest rating within our DB, right now they're rated based on the score between 0-1000, but for governance the highest score in our DB is about 500.
- Links in the info buttons not rendering as links (they function, just don't show up as blue and underlined). I don't think this will take too long to fix, however I couldn't find a quick fix for this and had to move on.
- These components/variables might be able to be condensed down into fewer files, as there is some repetition of code. This was not a priority, the priority was to get it working fully first.
- The SVG paths could probably be refactored out into a separate file and then all called on each of the components.

# Other Comments
Security concerns with links "target='blank'": https://www.freecodecamp.org/news/how-to-use-html-to-open-link-in-new-tab/

# React Console Screenshot
image


caolan-power (commented):
Looks good. No errors on my end and code looks good too.

One comment. Where did you get the svg stuff? do you need to credit it?

BearachB:
> One comment. Where did you get the svg stuff? do you need to credit it?

Very good point, I had entirely forgotten to comment this in. I've added the source to each of the three svg paths in as a comment. They all came from https://www.svgrepo.com/

caolan-power (approved):
Looks good after merge too

WarrenK-design (changes requested):
If that spacing is quick fix. 
image

Very good spot @WarrenK-design had that in earlier when I was testing the layout. Fixed now in latest push. Apologies

WarrenK-design:
Previous comment  fix
image

WarrenK-design (approved):
Really good work @BearachB these look a lot easier to gauge ESG at quick glance, I like the info button explanation too.

PR #435, opened 2022-10-27
Started on create portfolios

beckskelly:
1. Made portfolio ticker cards - they go on the page when the user has a portfolio
2. made create portfolio page - links from my portfolios page
3. made "my portfolios page" links in with nav bar

Note that half of "my portfolios" page is commented out. This is because the code left in makes errors, as its not being used yet, but we will need it down the line. The page has 2x divs, one is shown when the user has a portfolio, and one is shown when they dont

image

image

image


BearachB (approved):
All looks decent, I know it's all going to change eventually but good to have a starting point for each of these pages. 

Remind me next week @beckskelly and I'll help you in setting up all the other frontend stuff! I think would be good if we just power through and put together the bones of the frontend of the site so that the backend guys can start populating it as they get stuff done bit by bit.

joeycor92 (changes requested):
Not sure you did a pull before committing as all the portfolio routes, models, controllers aren't present. Other than that it's very good. Happy with it as a starting point.

beckskelly:
Good spot @joeycor92 , I was 43 commits behind!!! should have sorted that now

joeycor92 (approved):

PR #439, opened 2022-10-30
Frontend UI misc pages 404 error page

BearachB:
# Changes
- Added a 404/Page not found page component.
- Consists of a page with an error message for the user, along with some links to other pages (home, stock discovery, registration, login). 
- Added this to the routes using the wildcard *, meaning that the user will see this page if they navigate to any page that doesn't exist.

image

# Limitations
- Doesn't work for the search page (instead you get the search results error) 
image

# Future Work/Revisions
- The links on the page will need to be revisited at some point once we flesh out more pages on the site. 
- Also will need to show different versions of the page depending on whether the user is logged in or not (this is a larger problem though as there are no protected routes yet?)


beckskelly:
I like it, and good work setting up a page thats standard with websites that we've overlooked. 

However do we need links on it? most 404 pages just say page not found. Would resolve the issue you have of protected routes/ signed in and out etc. We have a navbar at the top. could always make it sticky if we wanted the site easier navigatable.

Also I think its fine that it doesn't work with the search page and only the error message shows up. From a useability point of view, if I searched for a garbage search term, I'd just want to search again, and not go to a 404 page.

Let me know your thoughts on the points above

BearachB:
> However do we need links on it? most 404 pages just say page not found. Would resolve the issue you have of protected routes/ signed in and out etc. We have a navbar at the top. could always make it sticky if we wanted the site easier navigatable.

Yeah could do it either way, maybe a question for the wider team. I don't think there's any downside for the minute in having the links on screen? Makes it clearer to the user where they can go if they do end up on the 404 page. Some sites will put in a search bar or just a link to the site homepage.

> Also I think its fine that it doesn't work with the search page and only the error message shows up. From a useability point of view, if I searched for a garbage search term, I'd just want to search again, and not go to a 404 page.

Yep I also think this is fine, just pointing out that it won't work on pages where the URL is parameterised.

BearachB:
@beckskelly What about just a button linking back to the homepage? Once we flesh out a splash page too it might look a bit snazzier (I think that's the technical term).

image

beckskelly:
I like that yes! Just make sure ye do a pull from master, and it says here there's conflicting files in App.js, so have a look at that before you merge

BearachB:
OK very weird, more Git weirdness happening. It has picked up some of your changes from your portfolio branch and has added them into my last push. This is bizarre because I made sure that they were not added or committed, however when I pushed they showed up? Need to figure out how to fix this, tried a few suggestions from StackOverflow but can't seem to remove them

BearachB:
Even weirder, when I try to rebase the branch back to before this commit, only commits from your branch are showing up.
image

BearachB:
OK finally solved it using this SO thread: https://stackoverflow.com/questions/40780587/remove-commits-from-an-existing-pull-request

Really confused as to how this happened in the first place as I was careful to only add or commit the 4 files I was working on.

beckskelly:
Fair play @BearachB , I like the home button for now. Wonder why Git has its own mind

beckskelly (approved):

BearachB:
> Fair play @BearachB , I like the home button for now. Wonder why Git has its own mind

Can always update it too! I moreso wanted to make the skeleton of the page so that we have it in place, so now updates to it will be fast. Would be good to have some fancy artwork etc.

There's some good examples in here: https://www.searchenginejournal.com/404-page-examples/211154/

caolan-power (approved):
Checked it out and all looking good. No errors and code looks good

PR #440, opened 2022-10-31
resolving error

beckskelly:
pushed an error from last thursdays work, just an unused import, sorry about this! should be all gone now

image


src/screens/myPortfolios/myPortfolios.js
```diff
@@ -1,6 +1,8 @@
-import { Container, Button, Row, Col } from "react-bootstrap";
+import { Container
+ //   , Button, Row, Col 
+} from "react-bootstrap";
```
BearachB, line 3: Is there a reason Button, Row and Col are commented out?

src/screens/myPortfolios/myPortfolios.js
```diff
@@ -1,6 +1,8 @@
...
 // import { useState } from "react";
-import PortfolioCard from "../../components/portfolioComponents/portfolioCard/PortfolioCard";
+// import PortfolioCard from "../../components/portfolioComponents/portfolioCard/PortfolioCard";
```
BearachB, line 5: Same here, do they need to stay commented out?

BearachB:
@beckskelly OK grand! What was the unused import that was included previously? I don't think I got an error when running it last week.

beckskelly:
> @beckskelly OK grand! What was the unused import that was included previously? I don't think I got an error when running it last week.

Row, Col, Button and the PortfolioCard

beckskelly:
> @beckskelly OK grand! What was the unused import that was included previously? I don't think I got an error when running it last week.

I think it changed heavily between reviews during the day which might explain why the errors slipped through!

BearachB (approved):
Grand OK, happy to approve for now. I'm not getting any errors when running it. Will be nice to see it once the back end is up and running.

caolan-power (approved):
Nice one. May look into removing that bootstrap country selector import too and remove it if we don't need it anymore

PR #446, opened 2022-11-01
Backend - Order Book - Order Schema

DaraghK93:
Small PR for the schema for the order book collection

Screenshot 2022-11-01 at 15 18 36

Next steps:
- create order route
- link to portfolio
- create lambda function that will run every 20 mins (when prices change) that checks what orders can be fulfilled and then actions the orders. this will probably use the same controller function that actions market orders


caolan-power (commented):
Cool so you're going to use the '_id' field generated from mongo as the orderID? then use that when full-filling orders for a portfolio is it?

PR #447, opened 2022-11-01
Machine learning news sentiment docker deploy

WarrenK-design:
Solution for deploying the larger machine learning lambda functions using Docker. The tutorial used is [here](https://aws.amazon.com/blogs/compute/using-container-image-support-for-aws-lambda-with-aws-sam/) its not actually that bad at all. 

Some small code changes from what I pushed last week but for most part it was a copy and paste over. The images are stored in Amazon ECR, when the lambda function is invoked it gets the image from the image repo. For the moment it is just the naive Bayes model deployed, going to go back and use Caolan F-Score functions to try and improve the model but it should only be a change to the pickle file within the ```MLData``` directory. 

Have changed parts of the default ```README.md``` that were not relevant but any questions ask me. This is the way we will need to deploy Twitter and News Recommendation system. 

Its not part of Amplify project anymore, it is its own thing deployed through the ```sam``` cli explained in the ```README.md```, need to find a flow for accepting deployments though maybe a review then you can deploy using ```sam deploy```?  The actual code for the lambda news sentiment function is now under ```machineLearning/sentimentAnalysis/newsSentimentAnalysis/NewsSentimentLambdaFunction```

If you do deploy new images too, delete old images from Amazon ECR to avoid storage cost going up.


BearachB (approved):
Looks really good @WarrenK-design, a lot of work in this! 

Comprehensive readmes included made it easy to understand where everything is coming from. However I'll definitely have questions for you when I'm deploying the recommender system. 

@caolan-power might have a better idea if there's any problems included in this PR but I can't spot anything out of place. Good work!

caolan-power (approved):
Ok so I did the tutorial and I'm understanding it now. Will try with my function and i'll get onto you if I run into any errors. 

Didn't need to check the python code as is the same as the lambda from last week.

Class stuff

PR #449, opened 2022-11-02
Machine learning news sentiment evaluating models

WarrenK-design:
Used the function @caolan-power developed to start generating reports for the following metrics: 
1. accuracy
2. positive precision
3. negative precision
4. positive recall
5. negative recall
6. positive f_measure
7. negative f_measure

The scripts for generating the models for News and Twitter sentiment will now write the results for these metrics for each model trained to a ```evaluationResults.csv``` file. 

Additionally another csv file called ```highestEvaluationResults.csv``` will be created with the highest results for each metric, these will also be printed to the console as shown below. 
image

@caolan-power only part of your ```train_model.py``` file was the evaluation section just make sure all still good, deleted the function you where using for generating report so we can both work of same one as discussed.


caolan-power (approved):
All good tested it and all working fine.

Much prefer the way it's printing and displaying now vs my dumping into a file.

BearachB (approved):
Good stuff, much clearer way of showing the evaluation results than before.

PR #450, opened 2022-11-02
Machine learning stock recommender system EDA

BearachB:
# Changes:
- This PR is to merge the starting point for the Recommender system.
- The main content of this is the Jupyter notebooks file "model1.ipynb". A Jupyter notebook was used as it's quick and easy to prototype ML type problems.
- There is a small amount of EDA in this notebook, however it ended up not exactly being needed.
- The main method of calculating similar companies is done using cosine similarity of the longbusinesssummary given in the original csv provided by @joeycor92. This is done by converting the words to vectors using the distilbert SentenceTransformer.
- The main function give_recommendations() takes an index input (i.e. a ticker symbol), and provides the top 20 similar ticker symbols as an output.
- This is a starting point for the recommender system, and the actual method of calculating similar companies will be adjusted down the line, once an established deployment is up and running. Some of the content from the notebook won't be needed in future (such as reading a user's stocks from the DB and posting back their recommendations), however I've left it in for posterity.

How this function works can be seen on the below plot with the 4 randomly selected companies. This plots cosine similarity against all 492 companies. For the first, GOOG (Alphabet) is the selected stock. As expected the closest to them is GOOGL (a different class of stock for the same company), followed by Apple Inc., Western Digital, Garmin, Amazon, Best Buy etc. This does a good job of grouping similar companies together, however it does not take into account any metrics other than the longbusinesssummary, so it does not account for size/ESG ratings/location etc.
image

# Next steps:
- Finalise converting this to a Python script (in progress).
  - Update function to take in multiple ticker symbols.
  - Add in try, except blocks with error checking into functions.
  - Add more comprehensive comments to functions.
  - Remove all unused imports (lots of imports were used to test various parts of this notebook).
- Create AWS Lambda function.
- Deploy lambda function using docker/SAM.
- Update stockRoutesService to invoke the recommender service every time a user visits the page. If they have stocks in a portfolio, recommend based on those; if they do not, recommend based on the top 10 biggest positive movers for the day (this step is subject to change).
- Clean up entire recommender system directory (bit messy right now as I'm settling into it.)
- Refine recommendation model using more inputs than just longbusinesssummary. This step might not even prove necessary as the results yielded at the moment seem to be realistic, but it will be revisited after successful deployment.


WarrenK-design (changes requested):
Hi Bearach, 

Looks good but having trouble running it, could you include your requirements.txt for installing dependencies? If you aren't using virtual environments have a look at this maybe https://docs.python.org/3/library/venv.html makes things a lot easier. 
Command to create a a vitrual environment with venv below. 
```
python -m venv Myenv
``` 
Activate that environment then and install dependencies the normal way. Then just do this command below to generate the requirements file. 
```
pip freeze > requirements.txt 
```
Within the readme too can you provide instructions on how to run what scripts/what the directory structure is? 

For the Python script too can you put this in a ```.env``` and dont commit it. 

image

Should just read it in from ```.env``` file then, have a look at the script ```main.py``` [here](https://github.com/DaraghK93/stockApp/tree/master/machineLearning/sentimentAnalysis/newsSentimentAnalysis/databaseUpdate) to see how to do it. 

Good start though.

BearachB:
@WarrenK-design OK finally addressed your comments, bogged me down more than I thought as I've never worked with venv before so had to do some reading to understand it.
For the running, the only file that should be run is the Jupyter notebook, you might need to install the Jupyter notebook extension for VS Code, from there you can just hit "run all" at the top of the notebook and it will execute all cells.
image

If you want to see it working, you can change the value of "stocks" for the user with the email: "recommender_test@gmail.com" in the database. The notebook will read in the ticker symbol you enter and will push the list of 20 recommendations for that ticker to the "recommended_stocks" field. As we discussed yesterday this won't be needed in future as it'll be run as a service, however I didn't want to delete it.
image

Any problems let me know!

Cheers,
Bearach

BearachB:
> Yeah I still actually cant get it to work unfortunately think just more to do with Juptyer/venv than anything, how are you telling Juptyer to use the the virtual environment version? 

Within the notebook I used the VS Code command palette (Ctrl + Shift + P) to select "Python Interpreter" and from there entered the path to the venv I had created `"\machineLearning\stockRecommendationSystem\venv\Scripts\python.exe"`. Apologies if this is not how it was meant to be selected, not familiar with the workflow. Might give you a call at some point to make sure I do it the correct way. 

> If just get rid of the mongoDB URI into .env I'll accept
Sorry I had removed it from the Jupyter notebook and put it into a dotenv but did not do that for the Python script. Changing now.

BearachB:
> Don't get bogged down on it though its not big deal I'll try run it when just converted to normal Python.

The Python script can actually be run, just the content isn't quite the same as the Jupyter notebook. It's hardcoded right now to return back 20 recommendations for the company with index 0 (3M), and will just print these as a list at the bottom of the page.

image

WarrenK-design (approved):
> Within the notebook I used the VS Code command palette (Ctrl + Shift + P) to select "Python Interpreter" and from there entered the path to the venv I had created \machineLearning\stockRecommendationSystem\venv\Scripts\python.exe. Apologies if this is not how it was meant to be selected, not familiar with the workflow. Might give you a call at some point to make sure I do it the correct way.

This fixed it yeah can run it now looks good.

Did get an error in the Pyhton script but I know you are still working on it.

image


Its to do with this, should be relative directory import the file is there but that path wont exist on everyone's machine use ```./sp500.csv``` to read it in and it runs ok.  

image

caolan-power (approved):
Hi Bearach very sorry for the delay in reviewing this. I'm bogged down in my own work so I forgot.

This is a cool first step and it's nice having recommendations working well off the longbusinesssummary.

Going forward I wonder could we add to the longbusinesssummary to make it 'live'.

Say we append the current daily range to each stocks, LBS. and/ or their E,S,G rating as a way of adding more inputs and it seems to work well textually. So something like 'Apple's environmental rating is rated POOR' it would be interesting to see if something like that would be possible. 

Well done though and great to be started on the recommendation system. Big part of the app

PR #452, opened 2022-11-02
Game - Leagues - League Schema

DaraghK93:
I've made a basic but easily scalable schema for leagues. I've made a createLeague and getLeagues route and controllers then to test this out, specifically to see if the populate method would join the user schema to it, which worked. created the league.model.js, leaguecontroller.js and leagueRoutes.js

if you want to test, send a post request to 'http://localhost:3001/api/league/createleague' and send a token with it.

Screenshot 2022-11-02 at 17 45 49

send a get to 'http://localhost:3001/api/league' to see the leagues created. it should show all the user's details with that. I can narrow down the exact details we need for displaying purposes when we know what we need.

To do:

- Make a join league route. this will involve first decoupling the portfolio schema from the user schema and just saving a reference to the portfolio ID in user. 
- finish order book schema, once the portfolio schema is ready


joeycor92 (approved):
Looks good and seems to work for me. The only thing, as mentioned in our chat, is that we might need to consider whether portfolios should be an input for createLeague. Other than that it looks good to me.

caolan-power (approved):
Looks good to me. 

So when a user creates a league they are automatically added to that league? If you've thought about it and the flow that's no bother. I more thought that the person who created it would be added to a 'creator' field or something which would give them admin access for moderation etc? Can talk through it more if you like (it'd only be a small change) but everything looks good so will approve

amplify/backend/function/expressServer/src/models/league.model.js
```diff
@@ -0,0 +1,18 @@
+const mongoose = require('mongoose')
+const {Schema} = mongoose
+
+const LeagueSchema = new mongoose.Schema(
+    {
+      leagueName: { type: String, trim: true, required: true },
+      budget: { type: String, trim: true, required: true},
```
WarrenK-design, line 7: Should this be a numerical value?

WarrenK-design (changes requested):

amplify/backend/function/expressServer/src/models/league.model.js
DaraghK93: That's a good spot. I'll change that now and update the file

DaraghK93:
@caolan-power, yeah, that's what I was thinking. Portoflios isn't required to be sent right now, so they won't be automatically added. there may be an option for them to do so n the frontend.

That's a good point, an admin field for the game creator could work

amplify/backend/function/expressServer/src/models/league.model.js
```diff
@@ -0,0 +1,18 @@
+const mongoose = require('mongoose')
+const {Schema} = mongoose
+
+const LeagueSchema = new mongoose.Schema(
```
WarrenK-design, line 4: For this too can you add in one more field called "Privacy" and should be either "public" or "private".

amplify/backend/function/expressServer/src/controllers/leagueController.js
```diff
@@ -0,0 +1,90 @@
...
+      // data is missing bad request
+      res.status(400)
+      res.errormessage = 'All details are needed to create a league'
+      return next(
+        new Error(
+          'The client has not sent the required information to create a league',
+        ),
+      )
+    } 
+    let league = await League.findOne({ leagueName })
```
WarrenK-design, line 36: See this too, I think the leagueName should not be unique, people will eventually call things the same name like two groups of different friends will probably have an "ESG League", the ID in database I think is enough to be unique.

WarrenK-design (changes requested):

amplify/backend/function/expressServer/src/controllers/leagueController.js
```diff
@@ -0,0 +1,90 @@
...
+    res.json({league})
+
+    } catch (err) {
+      console.error(err.message);
+      res.errormessage = 'Server error creating a league';
+      return next(err);
+    }
+}
+
+const getLeagues = async (req, res, next) => {
```
WarrenK-design, line 65: This a big one, I sent a request with no token and was able to get everyone's leagues which included there emails and passwords. Needs to be a protected route, I should only get the leagues I am associated with, should not return users details. 

Before changing anything on this route, is this the route which will be called when a user clicks into "Leagues" on the frontend is it?

image

WarrenK-design (changes requested):

DaraghK93:
Ok, added Warren's changes in there: "private" field, took uniqueness check out from leagueName, and have changed getLeagues to getPublicLeagues. this checks if it's private and if not returns the league - with no user details explicitly showm. Also auth middleware added to this route

amplify/backend/function/expressServer/src/controllers/leagueController.js
```diff
@@ -0,0 +1,89 @@
...
+    
+    // check if there are leagues
+     if (!leagues) {
+      res.status(404)
+      res.errormessage = 'No leagues found'
+      return next(new Error('No leagues found'))
+    }
+
+    //return the leagues
+    res.json({leagues})
```
WarrenK-design, line 77: Hi Daragh, 

Looks good just small thing in this to think about, returning user database ID's here too for league admin, think the admin field should be left out for now. 

Later on when a user wants to do any admin things we can check on back end and just store a "true"/"flase" for isAdmin on frontend, going to approve but can you remove this before you merge? 

image

WarrenK-design (approved):

joeycor92 (changes requested):
As discussed, focusable selector probably needs to be removed.
image

joeycor92 (approved):
Looks good now thanks.

DaraghK93:
Thanks all, all changes implemented

PR #453, opened 2022-11-03
Update longbusinesssummary in stock.model so that the information but…

joeycor92:
As mentioned in the chat, just updating the longbussinesssummary field in the stock model.

Data seems to be coming through ok now.

image


DaraghK93 (approved):
good stuff, glad we found why that wasn't coming through

beckskelly:
Thanks for sorting that @joeycor92

BearachB (approved):
Thanks for that @joeycor92.

Happy to approve however I'm not sure that the package-lock should be updated? The error you were getting was to do with the package I used to create the ESG ratings. This is working on the live site currently.
This may need to be reverted if this causes build issues when merged.

PR #454, opened 2022-11-03
Classifying tweets in db

caolan-power:
Classifying tweets already in DB. The script uses functions from the previous work to clean and classify the tweets

This method now should work for any brand new tweets found in the wild, not just the datasets I had for training (A few edge cases missed).


caolan-power:
I never checked format changes but it's weird that they are there on my files since I always have it on for my own code but unfortunately they are there i'll fix before adding anyone to avoid the pain

caolan-power:
Ok train_model.py is my file and there are no changes in it so don't worry about it. Weird that the formatter still had something to say about it considering i wrote it and formatted before

machineLearning/sentimentAnalysis/twitterSentimentAnalysis/databaseUpdate/README.md
```diff
@@ -0,0 +1,13 @@
...
+The script `main.py` is used to update the tweets already in the database
+
+## Choosing model
+
+The script expects a `.env` file within this directory.
+In the `.env` file are 3 parameters.
+
+1. MODEL_DIR - The base directory where the model is trained
+2. MODEL_NAME - The name of the model e.g. NaiveBayesClassifier
+3. MONGO_URI - The Mongo uri for the database to update.
```
WarrenK-design, line 13: All good, if you want to add just a section under this what model is being used now might be good idea actually just in case anyone runs it and overwrites it, might do same for my readme.

WarrenK-design (approved):

caolan-power:
good idea actually

BearachB (approved):
LGTM

PR #455, opened 2022-11-03
Update user model

DaraghK93:
I haven't attached to an issue as it's more of an ad hoc change so we can get the league/portfolio stuff going

Screenshot 2022-11-03 at 12 27 51

just added the references to these IDs (league will exist once my other PR is approved)

Also removed 'dob' field as 'overeighteen' covers that


BearachB (approved):
Good stuff @DaraghK93. 

I think spending the time now making sure the schemas are right and thinking through the relations of all the models will be worth it in the end.

PR #456, opened 2022-11-03
Backend update portfolio schema

joeycor92:
Just a quick update to the portfolio schema in line with Daragh's updates on the league schema and user schema. The routes will need to be updated in the next PR which will be to create a create portfolio service.

image

Not getting any errors.

### DIFF
--- a/amplify/backend/function/expressServer/src/controllers/portfolioController.js
+++ b/amplify/backend/function/expressServer/src/controllers/portfolioController.js
@@ -22,6 +22,7 @@ const createPortfolio = async (req, res, next) => {
         dateCreated: dateCreated
       });
       await User.updateOne({ _id: req.user.id }, {$push: {portfolios: newPortfolio}})
+      await newPortfolio.save()
         
       res.json({ newPortfolio });
     } catch (err) {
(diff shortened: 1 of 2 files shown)


BearachB (approved):
That seems to simplify it a bit! good work.

I'll be needing to access the user's holdings when I'm querying the recommender system service so good to see a simpler Mongo model for it.

DaraghK93 (changes requested):
Looks good Joey, just one thing onthe valueHistory field. we discussed a CRON job getting all the end of day values through calculating the holdings + remainder, so i think it should just be
valueHistory: [{date, endOfDayValue}]

joeycor92:
> Looks good Joey, just one thing onthe valueHistory field. we discussed a CRON job getting all the end of day values through calculating the holdings + remainder, so i think it should just be valueHistory: [{date, endOfDayValue}]

Good call. As discussed this will just be the total value updated at the end of each day.

joeycor92:
Updated now to remove 'required' for totalValue as this is a calculated field.

DaraghK93 (approved):
looks great, thanks for the change

PR #457, opened 2022-11-03
Fronend UI - Trade Page

beckskelly:
Finished off the UI for the trade page. Note there is one big glaring bug, the amount selected in the slider component. I will have to figure out how to round it down to 2 decimal points, have figured it out for displaying values, but not for that one as its input box is also configurable.

Work done is as follows:
1. Buttons for order type stay selected when clicked now, should hopefully infer they are "options" as per the feedback from the user evaluation
2. Fixed a small bug with slider/ input box where input box was not updating
3. Added a new display balance component
4. Added order summary component
5. Also tidied up a lot of the SCSS, hence all the deletions, made it more efficient

Also added a % symbol to individual stock page as per a change in the backend. Have highlighted the bug in the final screenshot.

Future work involves 
1. making the desktop look a bit better
2. changing the quantity component for when "Limit Order" is selected.
3. Adding an input box for "quantity" same as amount selected

image

image

image

image


DaraghK93:
Hi Becks, really good work, It's looking great. I think the way it is it should be easy to link up to backend as well. I like the slider as well and the buttons changing colour when pressed is great

DaraghK93 (approved):

beckskelly:
UPDATE

Managed to fix the bug with some head bashing and chats with @joeycor92 

image

joeycor92 (approved):
Great stuff. You're giving me more credit than I deserve there. Works brilliantly now, fair play to you.

PR #468, opened 2022-11-04
Database secure credentials 

WarrenK-design:
The database credentials in the Express app are now retrieved the same way as in the other lambda functions through AWS parameter store. 

If the value of the env variable ```ENVIRONMENT``` is ```prod``` then it will retrieve the production database credentials from AWS parameter store. 

If the value of the env variable ```ENVIRONMENT``` is ```dev``` then it will retrieve the database credentials from a local ```.env``` file. 

We all need to have a ```.env``` file under ```src``` with the contents below, have added this info to ```README.md``` too. 

```
ENVIRONMENT=dev
MONGO_URI=mongodb://localhost:27017/dev
JWT_SECRET=$Ckpc6ZlGF*Z3vJ89Nr
```


joeycor92 (approved):
Looks good. It's ok if jwtSecret is visible?

DaraghK93 (approved):
looks good, thanks for changing this. a lot more secure now

caolan-power (approved):
Nice one. Looks good to me

WarrenK-design:
> Looks good. It's ok if jwtSecret is visible?


Few things online have said it doesn't matter too much but before this is merged I may do the same to the JWT just to finalise all this security.

BearachB (approved):
All looks good @WarrenK-design. Thanks for this, definitely needed to tighten up security wise!

joeycor92:
> > Looks good. It's ok if jwtSecret is visible?
> 
> Few things online have said it doesn't matter too much but before this is merged I may do the same to the JWT just to finalise all this security.

Thanks for updating. Extra secure now!

PR #469, opened 2022-11-07
Machine learning news sentiment model training

WarrenK-design:
Final trained model included in this commit which has been deloyed to AWS SAM. 

The model is an MLPClassifier with the results of it shown below. The model was trained in iteration 5, over sampling was used in this iteration to eliminate the problem of data skewed towards neutral sentiment. 
image

This model was chosen due to its accuracy which was the highest, models highest in other categories are shown below which could be alternatives in the future. 
image

@caolan-power there are two functions which you may find useful in the ```modelFunctions.py``` file called ```upsampleTrainData``` and ```underSampleTrainData```, they used the process described [here](https://towardsdatascience.com/what-to-do-when-your-classification-dataset-is-imbalanced-6af031b12a36). 

The database name issue for the newsSentiment function is also fixed in this PR as done on the call on Friday.


BearachB (approved):
Apologies for the delay in getting this reviewed, kept getting side-tracked on the recommender system. 

All looks good as before @WarrenK-design! Lot of work here and good to see even more iterations of the model. Is this going to the be the last work carried out on the News sentiment? Or is there going to be more updates to it?

caolan-power (approved):
Cool stuff. I will use those sampling functions to try combat the skewed data in my models. Good work here

PR #479, opened 2022-11-07
Update the code for change in source data. Alpha Vantage had 4.close …

joeycor92:
So the data that comes in now saves the data in a different way so we need to make sure that the database is correct. I search all of the files for "4. close" and did a replace all to replace with "Close".

The database has changed from this:
image

To this: 
image

The data seems to be filtering through correctly as well.
image

As well as this, we only have 5 years of data.

Thanks


DaraghK93 (approved):
looks good, thanks for changing

caolan-power (approved):
Looks good

PR #481, opened 2022-11-07
Frontend user login check

WarrenK-design:
Fixed the routes that should be restricted for the site. Moved the logic to check if a user is logged out of the screens and into the App.js file. 

If the ```userInfo``` redux state is null this means a user is not logged into the application, for adding new routes do them in the form below. 

```
<Route path="/protectedRoute"
            element={userInfo? <protectedComponent/> : <Navigate to="/login"/>}/>
``` 
Fixed the Nav bar too, we had 2 buttons which went to the ```/``` route. The users name was also apperaing in black when logged in, was hard to see aginst the blue, changed it to white and text-font bold on hover. 

image
image

The route ```/``` now renders the ```HomeScreen``` component, nothing really on this page at the moment just there as placeholder, will fill this out when we can explain final system bit more, ```Play Now``` button takes you to register screen. 
image


DaraghK93 (approved):
looks Good Warren. all working on my end. I like that we are just protecting all functional routes now as well, makes it easier and forces people to sign up

beckskelly (approved):
Works perfectly, thanks for the work on this, app is a lot more standard feeling now, and nice one fixing those small issues with the navbar,

PR #483, opened 2022-11-07
Stock discovery overhaul

beckskelly:
as per discussion in discord - 

- removed most of the information off the cards
- tighened up space between rows and cards
- changed the side scrolling CSS, as I found a CSS solution that used less CSS and did not interfere with the card CSS
- got rid of right hand side scroller on page
- added in new side scrollers for desktop, unfortunately the desktop does not scroll/ work without them. Theyre gone in the mobile version
- got rid of background of card container
- moved daily change to bottom of card
- the cards are the same size regardless of on phone or in desktop.
- added a little chevron to the title copying the netflix example

I am aware some of the logos are absolutely TINY now but this is because the logos are all annoying and different sizes. Some of them are squares and some of them are rectangles so i went for a solution thats small and still shows the entireity of each logo. Let me know if ive gone too far and made them too small. The cards will have to get bigger however if we want bigger logos.

image
image

Let me know any issues


joeycor92 (changes requested):
So it looks great just a few things.

Still think that the dollar amount should always be 2 decimal and consistent. The percentage increase should be consistent as well, either one or two, I don't mind which.
image

I think the lack of scroller on desktop is a bit of a disaster. Like even at half screen it doesn't work. I don't think we can have that.
image

Just wondering as well, and might be a discussion for people, I think the company name should be on top and the ticker should be underneath. Think it looks a bit weird with just a D over the name. This one not that important but a question for the group.

image

beckskelly:
Fixed the decimal point, thanks for the spot. On the scrollers, they should be up in the desktop, I agree with your points on them as the desktop is unusable without them. I can see them on my end anyway. Could you pull and check again?

joeycor92:
> @joeycor92 Fixed the decimal point, thanks for the spot. On the scrollers, they should be up in the desktop, I agree with your points on them as the desktop is unusable without them. I can see them on my end anyway. Could you pull and check again?
> 
> Just changed the colour of them there to lighter as I felt the dark was too dark
> 
> image

No I mean that I can see the scrollers when I have it on full screen but I have it on half screen and they are gone. Like this is half my screen right now and there's no scroller. I've pulled there now so can see your changes.
image

Thanks for fixing the decimal. I had edited the comment above as wasn't sure you had seen yet just to note the -$ amounts are showing as $-

joeycor92 (approved):
Happy with it now. Although I do think we should think about the placement of the ticker symbols as the name is more well known to people.

src/components/stockDiscoveryComponents/tickercard/Tickercard.js
```diff
@@ -1,68 +1,55 @@
...
-        >
-          <ListGroupItem >
-            <h5 style={{ float: "left" }}>{stock.longname}</h5>
-          </ListGroupItem>
-        </ListGroup>
-        <Container style={{ height: "11.25rem" }}>
+      <Card className="h-100 tickercardstyle" style={{
+        padding: "none"
+      }}>
+        <Container style={{ height: "5.625rem", width: "7rem" }}>
```
WarrenK-design, line 31: See all this inline style stuff, taught we where using scss and classes?

src/components/stockDiscoveryComponents/stockSummary/stockSummary.js
```diff
@@ -31,19 +32,19 @@ function StockSummary() {
         <>
         {loading ? <LoadingSpinner /> : error  ? <MessageAlert variant='danger'>{error}</MessageAlert> :
         <>
-        <h3 className="stockdiscoveryRow">Today's Biggest Positive Movers</h3>
+        <h3 className="stockdiscoveryRow">Today's Biggest Positive Movers<ChevronRight /></h3>
```
WarrenK-design, line 35: These confused me a bit when I seen them on frontend I taught the heading was now clickable and the arrow was saying to go somewhere. I think they will get that bar at bottom means scroll. 
image

It also looks bit odd on mobile I think. 
image

WarrenK-design (changes requested):
Good work Becks, look a lot cleaner, few small things.

beckskelly:
@WarrenK-design got rid of the chevrons, agreed theyre too confusing. Also got rid of most of the inline styling (some of the bootstrap elements didnt accept my new classes so had to leave inline)

I get the point on inline styling and have extracted them to external classes just it feels a bit redundant as I think theyre genuinely not going to be used anywhere else? also each individual element is styled uniquely so I couldnt reuse any of the classes within the element

WarrenK-design (approved):
Hi Becks, 

Looks good! 

Yeah I get you I know for small things the inline styles are handy alright and sometimes bootstrap wont take the custom classes/bit finicky so I suppose if using classes for majority that's grand. Yeah I get the reusability too but at same time also even if it isn't reused handy to change it in class but yeah just something that caught my eye think what you have now is grand. 

See if bootstrap isn't taking you css styling from a stylesheet, try using this [here](https://www.w3schools.com/css/css_important.asp) 

In stylesheet it would be:
```
.myClass{
  cssAttribute: Value !important; 
}
```
Helped me get around it a few times but doesn't always work.

PR #484, opened 2022-11-08
Backend update portfolio routes controller

joeycor92:
Create portfolio service. We don't have it set up to work with the league yet but at the moment a user can create a portfolio and we will see a reference in the user schema as well. 
image
image
So let me know if this looks ok.


WarrenK-design (changes requested):
Could you throw in a route in the API spec, not 100% sure what route to test/what data to send, cheers.

joeycor92:
> Could you throw in a route in the API spec, not 100% sure what route to test/what data to send, cheers.

Yes sorry,

So you're going to need a token. I'm actually just after realising that I have a console.log(token) in there that needs to be taken out, but you can leave it in there for now so that you can get one.

You'll need to log in to get the token.

Then you'll send a request to 
http://localhost:3001/api/user/createPortfolio

With the body as:
{
    "portfolioName": "My portfolio name",
    "startingBalance": a number
}

The starting balance will come from the league when that's ready.

amplify/backend/function/expressServer/src/controllers/portfolioController.js
```diff
@@ -1,30 +1,23 @@
...
-      dateCreated = new Date()
-      if (!req.body.remainder){
-        remainder = 10000
-      }
-      else {
-        remainder = req.body.remainder
-      }
-      const newPortfolio = new Portfolio({
+    // create object to be used in createPortfolio service
+      const portfolioData = {
```
WarrenK-design, line 14: We need to have a check here for required fields. 

When I send no data in body the app crashes. 
image
image

When I send only a portflioName I can create it but it has no balance. 
image
image

When I send a balance only it returns a 200 status but no portfolio is created? 
image
image

If I send wrong data type app crashes. 
image
image

amplify/backend/function/expressServer/src/controllers/portfolioController.js
```diff
@@ -1,30 +1,23 @@
...
-        dateCreated: dateCreated
-      });
-      await User.updateOne({ _id: req.user.id }, {$push: {portfolios: newPortfolio}})
-      await newPortfolio.save()
-        
-      res.json({ newPortfolio });
+        startingBalance: req.body.startingBalance,
+        userId: req.user.id
+      }
+      const createdPortfolio = PortfolioService.createPortfolio(portfolioData)
```
WarrenK-design, line 19: Need to await this statement its async, the return will always be empty, should be the newly created portfolio object.

amplify/backend/function/expressServer/src/controllers/portfolioController.js
```diff
@@ -1,30 +1,23 @@
...
+      res.json(createdPortfolio);
```
WarrenK-design, line 20: This will always be empty, update in excel too shouldn't be empty should be the portfolio object.

WarrenK-design (changes requested):
Hi Joey, good start but getting good few errors if you could have a look into them.

joeycor92:
> Hi Joey, good start but getting good few errors if you could have a look into them.

All those changes have been made I think with the excel updated as well.

amplify/backend/function/expressServer/src/controllers/portfolioController.js
```diff
@@ -1,30 +1,44 @@
...
 // @access private
 
-const createPortfolio = async (req, res, next) => {
+const createHangingPortfolio = async (req, res, next) => {
   try {
-      dateCreated = new Date()
-      if (!req.body.remainder){
-        remainder = 10000
+    // create object to be used in createPortfolio service
+      console.log(parseFloat(req.body.startingBalance))
```
WarrenK-design, line 14: Delete this console log

WarrenK-design (approved):
Hi Joey, 

Looks good no errors now just delete the console log before you commit. 

Just maybe one for future too I see we have ```remainder``` in the ```portfolio```, will we save ```startingValue``` as well? Might save us down the line for calculating percentage gain/loss instead of a big math function for all the transactions.

DaraghK93:
Hi Joey, loads of really good work here. works great. we should be able to link up our work very soon. the only thing I'm wondering about here are the routes laid out in app.js. portfolio seems to still be in user. not a massive issue
Screenshot 2022-11-08 at 17 13 09

DaraghK93 (approved):

WarrenK-design:
> Hi Joey, loads of really good work here. works great. we should be able to link up our work very soon. the only thing I'm wondering about here are the routes laid out in app.js. portfolio seems to still be in user. not a massive issue Screenshot 2022-11-08 at 17 13 09

Yeah I noticed that too alright, I'd say portfolio could nearly be its own thing

PR #489, opened 2022-11-09
Backend - Game - Join League/Create League

DaraghK93:
A fairly large PR here, apologies for that. 3 main things:

League Schema: this should be updated as per our meeting at the beginning of the week. I have also included the "finished" field.

createLeague:
I've amended this since my last PR so maybe test it with the new schema and slightly different controller.
{
    "leagueName": "league16",
    "startingBalance": 2229000,
    "leagueType": "valueBased",
    "tradingFee": 24,
    "private": false,
    "active":true,
    "finished": true,
    "startDate": "2022-12-24",
    "WinningValue": 200002,
    "maxDailyTrades": 14
}

The controller calls the leagueServices to generate a random 6 character string for the access code. the user who sends this is the leagueAdmin and their Id is added to this field. 

JoinLeaguebyCode:

this is when a user sends the request just with an access code. to test this just create a league and grab the accesscode and send it like this

{
    "accessCode": "6ggh8t"
}

this function then creates a portfolio, adding the leagueId and userId to the portfolio through the service. the portfolioID and userID is then pushed to the league object, lastly the leagueID is added to the user object.

The routes are  POST to /api/league/createleague and POST to  /api/league/joinleague

NOTE: you must send a token in the header with each of these requests


WarrenK-design (changes requested):
Hi Daragh, 

Good work, few things. 

# Schema 
Can you add an image field to the league schema it will be a string. 

# Data sent
In addition to what you have to create a game I should send. 

1. Image - Will just be a string. 
2. minERating - Optional default to highest 
3. minGRating - Optional default to highest 
4. minSRating - Optional default to highest 
5. Sectors - Optional list of sectors, default to all.
6. EndDate - Optional based upon the game type 

# Validation 
Can see just a few things here that will break it. 
1. I can create a league thats already finished. 
2. I can create a league that starts today but is not active.
3. I can create a league with a budget of 0. 
4. I can create a league with winning value of 0. 
5. Can create league with a start date in the past. 

Should be a check in backend for things like that but also, things like ```active``` dont rely on user for that we have logic to do in backend if they supply the ```startDate```. 

# Return
Should that users have something in it? Same with portfolios. 
Might be handy to return json object 
```
{
 game: {gameObj},
portfolio: {MyPorfoflio}
} 
```
Can see @beckskelly needing the game and current portfolio when we go from ```createGamePage``` -> ```individualGamePage```
image

amplify/backend/function/expressServer/src/controllers/leagueController.js
```diff
@@ -11,20 +11,31 @@ const createLeague = async (req, res, next) => {
...
             leagueName,
-            budget,
+            startingBalance,
             leagueType,
             tradingFee,
-            portfolios,
-            private
+            private,
+            startDate,
+            WinningValue,
```
WarrenK-design, line 19: Should this be lower case? Just will be good to have unified so doesn't trip us up when sending data remembering that this is upper case.

amplify/backend/function/expressServer/src/controllers/leagueController.js
```diff
@@ -36,29 +47,137 @@ const createLeague = async (req, res, next) => {
...
+        res.errormessage = 'time based games must be at least a day long'
+        return next(
+          new Error(
+            'time based games must be at least a day long',
+          ),
+        )
+      }
+    }
+
+    if (tradingFee < 0) {
```
WarrenK-design, line 121: Can we do a trading fee of 0? Like a free for all league this just general question.

joeycor92 (changes requested):
Nearly there I think. As mentioned in chat, we should probably have an enum for images as we don't want them submitting a random image.

As well as this, we discussed that WinningValue shouldn't be a requirement if the timeBased game is chosen. Will keep looking but just the first two things I've noticed! Thanks

amplify/backend/function/expressServer/src/controllers/leagueController.js
```diff
@@ -36,29 +47,137 @@ const createLeague = async (req, res, next) => {
...
+      res.status(400)
+      res.errormessage = 'Start date is in the past'
+      return next(
+        new Error(
+          'Start date must be in the future',
+        ),
+      )
+  } 
+    let active; 
+    if (start == today) {
```
WarrenK-design, line 88: This logic took be a min might be better to write.

```
if (new Date(start) < today) {
   <error the league is in the furture>
}else if (start == today){
   <set active to true we are starting today>
}else{
   <set active to false we have a scheduled league> 
}
```

amplify/backend/function/expressServer/src/controllers/leagueController.js
```diff
@@ -82,8 +201,67 @@ const getPublicLeagues = async (req, res, next) => {
...
+      let league = await League.findOne({ accessCode })
+
+      // 404 for no such league
+      if (!league) {
+        res.status(404)
+        res.errormessage = 'Invalid Access Code'
+        return next(new Error('Invalid Access Code'))
+      }
+
+      console.log(league)
```
WarrenK-design, line 240: Get rid of this

amplify/backend/function/expressServer/src/controllers/leagueController.js
```diff
@@ -82,8 +201,67 @@ const getPublicLeagues = async (req, res, next) => {
...
+  try {
+
+      // get the accessCode
+      const {accessCode} = req.body
+      if (
+        typeof accessCode === 'undefined'
+      ) {
+        // data is missing bad request
+        res.status(400)
+        res.errormessage = 'an access code is needed'
```
WarrenK-design, line 220: Capital letter for this will be shown to user

WarrenK-design (changes requested):
Hi Daragh, its nearly there just a few comments for small things then I'll give it another test.

amplify/backend/function/expressServer/src/controllers/leagueController.js
joeycor92: > Can we do a trading fee of 0? Like a free for all league this just general question.

Personally think we should but there should be a warning that it's completely ridiculous and unrealistic.

joeycor92 (approved):
Ok seems to be working for me. Have tried a few different inputs and all seem to work.

WarrenK-design (changes requested):
Hi Daragh, 

I'm sorry for this its nearly there, two small errors then just some edge cases that just make note of we will get to them in testing.

When I create a timeBased game I can still set the winning value, might cause errors later on if we do any queries with winning value. 

image

I have tried out join league too, I have created a league with 1 user and then joined it with another, I get back the league object but the ```users``` is still empty? Would expect there be two users in this object (creator and new person that just joined)? 
image



### Edge Cases  
1. Sending empty sectors array creates a league with no sectors available to trade. 
2. Sending an image that isn't in the enum causes a server error.

WarrenK-design (approved):
Fair play Daragh this is good chunk of work! Well done

PR #491, opened 2022-11-09
Backend buy stock

joeycor92:
Right folks this is a big one.

Holdings Schema has been created in line with the ERD.
image

A buy stock service has been created which will be reused in the limit orders.
A buy stock controller function has been created, with a route api/portfolio/buyStock
The route for creating a portfolio has also been changes from api/user/createPortfolio to api/portfolio/createPortfolio

I've added comments in to explain how this is all done. There are a lot of checks.

This is a postman request
image

It returns the updated portfolio.

Below is how the DB looks for portfolio.
image

Holdings: 
image

Transactions:
image


DaraghK93 (changes requested):
Hey Joey, great work here so far. just a few things from me. first you're letting the user send down the "value" field. this should be calculated from the information about the stock and the units.

also, I am getting a server error on the first time i run it on a certain portfolioId. 
Screenshot 2022-11-10 at 09 48 21

the missing x-api gateway messages are the successful ones in between and the second server error is when I changed portfolioId. I think it might be an async issue with newRemainder but can't be sure.

also I think that this should be a protected route, so add in that here, like the one above

Screenshot 2022-11-10 at 09 41 27

this may require an additional check then for the userID that you get from the token. you would need to check that the userID is contained in the portfolio, as otherwise you could buy stocks for other people's portfolios

joeycor92:
> Hey Joey, great work here so far. just a few things from me. first you're letting the user send down the "value" field. this should be calculated from the information about the stock and the units.
> 
> also, I am getting a server error on the first time i run it on a certain portfolioId. Screenshot 2022-11-10 at 09 48 21
> 
> the missing x-api gateway messages are the successful ones in between and the second server error is when I changed portfolioId. I think it might be an async issue with newRemainder but can't be sure.
> 
> also I think that this should be a protected route, so add in that here, like the one above
> 
> Screenshot 2022-11-10 at 09 41 27
> 
> this may require an additional check then for the userID that you get from the token. you would need to check that the userID is contained in the portfolio, as otherwise you could buy stocks for other people's portfolios

Thanks Daragh, great spots.

Have updated for all of those changes if you'd like to have another go.

DaraghK93:
Just one thing, when I put the wrong stockID in, it doesn't send back a 404 like you have written, it's trying to send it to the db and can't find that objectID, so sends a 200 response and server error message. 

Also, your buy stock service should be in a try/catch block, as you're making a DB call in there

joeycor92:
> Just one thing, when I put the wrong stockID in, it doesn't send back a 404 like you have written, it's trying to send it to the db and can't find that objectID, so sends a 200 response and server error message.
> 
> Also, your buy stock service should be in a try/catch block, as you're making a DB call in there

Updated now so that it checks and added in a try/catch block! Good spots.

joeycor92:
Sample request

portfolio/buyStock
{
    "portfolioId": ,
    "stockId": "63457b65e940127420948c8e",
    "units": 5,
    "buyOrSell": "BUY",
    "orderType": "MARKET"
}

Headers needs a token as well.

Won't work if no portfolio created.

DaraghK93:
Thanks for making the changes Joey. few things:

same thing is happening for the portfolio Id that happened for the stockId

Screenshot 2022-11-10 at 19 24 12
you're console logging the stock here so remove

also need a check on ordertype on the backend as i put in a different value and the mongoose error is coming up, think it should just be a 400.

DaraghK93:
also noticed this.  if an error happens in the buyStock service, you've passed it to the error middleware. we don't have access to that in the services. you only have access to req,res,next in the controllers. you'll need just throw a new error and the controller will pick it up. take a look at my leagueServices, especially join league
Screenshot 2022-11-10 at 19 29 56

joeycor92:
> also noticed this. if an error happens in the buyStock service, you've passed it to the error middleware. we don't have access to that in the services. you only have access to req,res,next in the controllers. you'll need just throw a new error and the controller will pick it up. take a look at my leagueServices, especially join league Screenshot 2022-11-10 at 19 29 56

Thanks for spotting those again. Fingers crossed....

DaraghK93:
just one last thing. this is saying invalid length, for portfolio Id as well. if i put in one that's the right length but isn't a stock id it still says invalid length. 
Screenshot 2022-11-11 at 09 51 05

instead of this i think it would be better to use mongoose to check.
import it at the top of the page
const mongoose = require('mongoose');

instead of the above if have:

if (mongoose.Types.ObjectId.isValid(stockId) === false ) {
     do error handling with 404 no stock found
}

this way it will catch the no stock found there, and if the id is correct but not in the database it will catch it in your other 404 block for stock === null. 

Same goes for PortfolioID

DaraghK93:
I think you should just change this to no stock found. for both stockId and PortfolioId, it's not just invalid length that this catches. can have the same error for these few because they're all just 404 no stock found.

Screenshot 2022-11-14 at 11 19 15


question on Limit Orders, I sent one and it deducted from my remainder and added the stock. should this happen or should there be a check? or do we want to just make this the market order route?

apart from these all good.

joeycor92:
> I think you should just change this to no stock found. for both stockId and PortfolioId, it's not just invalid length that this catches. can have the same error for these few because they're all just 404 no stock found.
> 
> Screenshot 2022-11-14 at 11 19 15
> 
> question on Limit Orders, I sent one and it deducted from my remainder and added the stock. should this happen or should there be a check? or do we want to just make this the market order route?
> 
> apart from these all good.

Thanks very much, have made those changes. The route is now 
api/portfolio/buyStockMarketOrder

It must be order type "MARKET" to actually work.

DaraghK93:
Thanks for all those changes Joey. Approving now. good work

DaraghK93 (approved):

amplify/backend/function/expressServer/src/models/holdings.model.js
```diff
@@ -0,0 +1,16 @@
+const mongoose = require('mongoose')
+const {Schema} = mongoose
+
+const HoldingsSchema = new mongoose.Schema(
```
caolan-power, line 4: Would an 'average price' field be worthwhile? Rather than calculating a users average cost by going through each and every historical transaction? Maybe not useful for now but for certain metrics like 'person with biggest percent gain' or biggest gain. Or if not it would be an easy metric to display to the user that they are down on 'amzn' overall for example. Helping them play the game better. Just something to think about

caolan-power (approved):
Hi sorry this took awhile.

All looks good and i'm approving but just putting down some thoughts as I go. Might be useless to you or already decided against since you've put more thought into this than I have.

Good error handling. I messed around with different requests, sending different data each time and there were clear messages as to what was wrong with each request.

Also, probably is a good thing having a market order as it's own thing for now but as we add rules and trading fees are added or restrictions etc. will the buy mechanism be the same for limit order and market order? one has just a delayed execution time (waiting for a certain price to be hit vs using currentprice). 
Will game rules be accessed within the portfolio services buyStock function to add fees, assign penalties etc. ? Therefore buyStockMarketOrder would just be execute order or something like that? but then again reusability within the services for fees etc is probably enough and easier to get it implemented. 

This probably has been thought about already as I said but I'm just thinking for reusability down the line. Food for thought if nothing else.

Great work! It's really cool seeing this functionality come together

PR #492, opened 2022-11-10
Portfolio page started

beckskelly:
Contents of this PR are:

1. Started the Portfolio Page
2. Portfolio overall balance component

Decoupled the chart from the button as per discussion in discord, good suggestion @WarrenK-design 

Reformatted a lot of the the stock price chart page code by accident, I was refactoring classes that were in common with the new components. so you can essentially ignore most of the code changes in src/components/stockVisualisationComponents/ChartTypes/PriceChart/PriceChart.js, its only a couple of classNames that were added in. 

Future work:
1. Dummy data in for everything, one part of the logic is not entirely perfect, like for changing the graphs using the dropdown. When we have portfolios coming through I'll fix this. In short I'm aware the portfolios attached to the user are an array of id's and I've programmed it as names for now.

2. The tooltip is irritating me a lot, as I've made a custom one I fixed a lot of formatting issues that were seen in the other graph for stock prices, however it has a border thats too thick in my opinion, I've tried getting rid of it, I'll have to sort it in the future.

3. Have a good idea how to format the Y axis as well so that the dollar sign is in front of the value and I'll be fixing this in the future

4. Will add in the color change for this graph as well (red/green), have to just think about the logic for it as the model is differnet to the stock model.

5. Also the "route" at the top of the page is wrong, its set to /portfolio for now but in the future it will be something like /game/:gameid

image
image


joeycor92 (commented):
Just wondering will we be having 1M, 1W, 1Y buttons for the graph?

beckskelly:
@joeycor92 it's a really good question, I decided that we weren't as none of the portfolios are going to be big enough for a year, month, etc. Now that theyre tied to the game as well, I very unsure what to do, its why I didn't just refactor the price graph from the stocks, also that graph is very tightly coupled with the information on the page, it felt like a headache trying to decouple although looking at it I think it could be easily done

joeycor92:
> @joeycor92 it's a really good question, I decided that we weren't as none of the portfolios are going to be big enough for a year, month, etc. Now that theyre tied to the game as well, I very unsure what to do, its why I didn't just refactor the price graph from the stocks, also that graph is very tightly coupled with the information on the page, it felt like a headache trying to decouple although looking at it now I think it could be easily done

I'd imagine some games will last up to a year, especially if a recession hits and they're value based games and suddenly everyone's value drops haha. However, I think you're right just to have a chart for since inception though, might be handier for the game chart. Happy with that

WarrenK-design (changes requested):
Hi Becks looks good, just small changed might save yourself some time, more so the tooltip one than anything.

src/components/portfolioComponents/portfolioGraph/portfoliograph.js
```diff
@@ -0,0 +1,68 @@
...
+        else {
+            setTickBoolean(false)
+        }
+    }
+    useEffect(() => {
+        showTick()
+    }, [])
+
+
+    const CustomToolTip = props => {
```
WarrenK-design, line 23: Break this tooltip out into its own component under ```widgets``` and require it. You can save yourself work here, the tooltip can then be reused for the individual stock page graph, comparison graph, and any other graphs which have axis of "Date" and "Price".

src/components/portfolioComponents/portfolioOverallViz/PortfolioOverallViz.js
```diff
@@ -0,0 +1,123 @@
...
+        { date: '01-23', value: 900 },
+        { date: '01-24', value: 100 },
+    ]
+    // simulates a list of portfolio names attached to user. Basic logic in but will have to change with real data.
+    const portfolioList = ["PortfolioA", "PortfolioB", "PortfolioC"]
+
+    const [data, setData] = useState(PortfolioA);
+
+    const handleSelect = (e) => {
+        if (e.target.id === "PortfolioA") {
```
WarrenK-design, line 67: For these I know its dummy data but just in general if you find yourself writing 2+ of same if statement try to break it into a loop, I think the portfolios will be in array more so too. 

Pseudo code of what I mean below. 
```
portfolios = [
    {"name":"portfolioA", "values":[{"date":"dateVal","value":"$"}, ....]},
    {"name":"portfolioB", "values":[{"date":"dateVal","value":"$"}, ....]},
    {"name":"portfolioC", "values":[{"date":"dateVal","value":"$"}, ....]}
]

for portfolio in portfolios{
   if e.target.id === portfolio.name{
     setYourState
    }
}
```

WarrenK-design (changes requested):
Hi Becks looks good, just small changed might save yourself some time, more so the tooltip one than anything.

WarrenK-design (approved):
Looks good!

joeycor92 (approved):

PR #497, opened 2022-11-10
Machine learning stock recommender system model testing

BearachB:
# Changes
- This PR contains most of the files needed to deploy the stock recommender system to the live application.
- The lambda function has been created, this now needs to be linked in with the stockRoutesService on the backend of the application (hoping to get this done tonight, however wanted this PR in).
- The function that was deployed is based on the model_2.py Python script. This has been hardcoded to return recommendations for the ticker symbol "AAPL" temporarily, this will be updated tonight to an input based on the user's holdings.
- dataexplore.py file was added so you can run all the EDA plots/prints easily (I think neither of you could use the Jupyter notebook in the last PR).

---

# Future Work/Next Steps

- Deploy lambda function using docker/SAM (currently SAM is running locally).
- Update stockRoutesService to invoke the recommender service every time a user visits the page. If they have stocks in a portfolio, recommend based on those; if they do not, recommend based on the top 10 biggest positive movers for the day (this step is subject to change).
- Refine recommendation model using more inputs than just longbusinesssummary. This step might not even prove necessary as the results yielded at the moment seem to be realistic, but it will be revisited after successful deployment.

---

# Screenshots 
## Temporarily hardcoded value of "AAPL"
image

## Lambda Returns the 20 recommended stocks for AAPL
image


caolan-power (approved):
Nice one. 

Built and ran it and all is working as expected. Didn't dive too much into the recommender system mechanics but enough to know the idea. I like it so far and think there are good options going forward with this. Great to have it set up in the SAM app now and it'll be plug and play with new models from now on.

machineLearning/stockRecommendationSystem/stockRecommender/events/event.json
```diff
@@ -0,0 +1,63 @@
+{
```
WarrenK-design, line 1: This isn't so much a change but I think this will make it a lot easier when testing a few different inputs for future work. 

When you call ```sam local invoke``` it defaults to this ```events.json``` file as the input which just has a bunch of dummy data in it now because its the default one. This can be used though to mock your API calls for different scenarios. 

In your Lambda function you can access this data through the ```event``` object passed as an input to your handler. This will contain the information in this file which is data which could be passed through an API call. 

You can create more than just one and use ```sam local invoke --event <PATH_TO_FILE>``` to change the event. So you might have a file which will pass a userId who has environment stocks called ```environment_event.json``` another maybe that has lots of social stocks in ```social_event.json``` and so on. Will make it handier to test and will replicate the deployment a bit more as it will be an API call in the end this is run on. 

Have a read [here](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-local-invoke.html)  and [here](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-concepts.html#gettingstarted-concepts-event) for bit more info on events.

WarrenK-design (changes requested):
Hi Bearach, 

Really good work built and ran successfully using ```sam build``` and ```sam local invoke```. 

The changes are just small things which make it bit easier in future, the events one dont worry about for now but will make future work lot easier. 

Was mentioned too quite a large PR, we all guilty of it last week smaller ones this week 👍

machineLearning/stockRecommendationSystem/stockRecommender/events/event.json
BearachB: That's really good information @WarrenK-design, thanks a million!

BearachB:
@WarrenK-design Addressed your comments. Really do appreciate the thoroughness! Added try catch statements to all along with Docstring comments.

WarrenK-design (approved):
Looks good Bearach!

PR #500, opened 2022-11-10
Game frontend - Create Game - Page 1 - Game Name, Type and Image

WarrenK-design:
I don't expect a review on this tonight I know its late, can demo on localhost tomorrow. 

This is first page in the create game flow.  User has to select a name (defaults to one already), a game type and a image.  Default names for issue we discussed about sanitising inputs on backend. 

The ```GameCreationOptionsCard``` component also has a new prop to disable the next button if all fields have not been completed. 

image
image
image
image

### DIFF
--- a/src/components/gameComponents/createGameScreenComponents/GameCreationOptionsCard.js
+++ b/src/components/gameComponents/createGameScreenComponents/GameCreationOptionsCard.js
@@ -5,11 +5,12 @@ import {Card,Row,Col,Button} from "react-bootstrap"
 
 /// GameCreationOptionsCard ///
 //  poprs:
-//      children  - The children of the component https://reactjs.org/docs/composition-vs-inheritance.html
-//      setScreen - Setter to set createGameScreen state
-//      screen    - The screen state set by setScreen
-//      gameType  - The type of game, depedning upon game type there will  be extra screen for valueBased game 
-function GameCreationOptionsCard({children, setScreen, screen, gameType}){
+//      children         - The children of the component https://reactjs.org/docs/composition-vs-inheritance.html
+//      setScreen        - Setter to set createGameScreen state
+//      screen           - The screen state set by setScreen
+//      gameType         - The type of game, depedning upon game type there will  be extra screen for valueBased game 
+//      disableNextStep  - This is boolean, true then cant go onto next step, false then you can go onto next step 
+function GameCreationOptionsCard({children, setScreen, screen, gameType, disableNextStep}){
     return(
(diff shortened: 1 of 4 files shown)


DaraghK93 (commented):
Hi Warren, looks great. really coming together. 

Just two things that worry me slightly. one being the drop down for names. it's a good work around for people calling the games ridiculous things. it may get a bit confusing if they have a few games called the same thing however. we've discussed having more options in the drop down, so this could work as well.

One main thing i noticed was that it's kind of hard to see which image is selected here
Screenshot 2022-11-10 at 20 08 04
not suer what the workaround is. maybe make the outline a little thicker? I found it quite hard to see at the start. also is there any point in telling them what the image they're picking is for, or do you think people will just guess that will be the game logo?

WarrenK-design:
> Hi Warren, looks great. really coming together.
> 
> Just two things that worry me slightly. one being the drop down for names. it's a good work around for people calling the games ridiculous things. it may get a bit confusing if they have a few games called the same thing however. we've discussed having more options in the drop down, so this could work as well.
> 
> One main thing i noticed was that it's kind of hard to see which image is selected here Screenshot 2022-11-10 at 20 08 04 not suer what the workaround is. maybe make the outline a little thicker? I found it quite hard to see at the start. also is there any point in telling them what the image they're picking is for, or do you think people will just guess that will be the game logo?

Hi Daragh yeah if we can come up with few more names or maybe a random name generator on backend route? For now though yeah might just be the workaround. 

Yeah I'm not sure I may leave it for now and we could see from evaluation if people get it or not, I've been looking at it too long to know if people will get it or not. 


Added a larger border to this. 
image

DaraghK93 (approved):
That looks a lot better, and will be a lot easier to spot.

I'll approve this, we can chat about the other things tomorrow/next week. great work

BearachB (approved):
100% I'd agree with @DaraghK93's concerns on the league names. Happy to go with your solution for the moment but need to have a think about how to implement it to give users some choice. Maybe just validate the names using a name library? Haven't looked for this but surely React has libraries for this exact scenario?

WarrenK-design:
> 100% I'd agree with @DaraghK93's concerns on the league names. Happy to go with your solution for the moment but need to have a think about how to implement it to give users some choice. Maybe just validate the names using a name library? Haven't looked for this but surely React has libraries for this exact scenario?

Grand job, yeah I do agree it is a problem alright but again something we could sink hours into. I think we put it on back burner for now anyway.

PR #504, opened 2022-11-14
Backend-Update- create-league

DaraghK93:
Very small PR here. I've just added a check into sectors. it will loop through the sectors sent until it finds a sector that doesn't exist in the schema, where it throws an error.
Screenshot 2022-11-14 at 11 46 43

Example request to test it:
{
    "leagueName": "league19",
    "startingBalance": 100000,
    "leagueType": "valueBased",
    "tradingFee": 3,
    "startDate": "2022-11-15",
    "winningValue": 1000000,
    "image": "/stock_photo_1.jpg",
    "maxDailyTrades": 3,
    "sectors": ["sdfsdf"]
}

can get rid of sectors then, or put in some valid ones from the array


caolan-power (approved):
Works as expected. LG2M

joeycor92 (approved):
Tried my best but couldn't break it. Good job.

PR #505, opened 2022-11-14
Game frontend create game page2 duration

WarrenK-design:
Second page of the create game flow, used to set duration. 

If the game is value based then only a start date required. 

image

If game is duration based needs a start and end date. The state has to be set to a ```Date``` object, start date defaults to today but end date is two weeks from today, user can change these if they want just starting values. 
image

Desktop views 
image
image

***Packge.json***
New module react-calendar has been installed but I have also removed that location picker pacakge as it was used nowehere and caused problems. 
Please check this branch out, do npm install and just make sure all still works fine.


BearachB (approved):
Looks really good @WarrenK-design. Good to see it coming along.
One thing that might be worth doing is making the "Back" and "Next" buttons slightly wider on all screens, at the moment they look a bit skinny to me and stand out (maybe that's just me though)

WarrenK-design:
> Looks really good @WarrenK-design. Good to see it coming along. One thing that might be worth doing is making the "Back" and "Next" buttons slightly wider on all screens, at the moment they look a bit skinny to me and stand out (maybe that's just me though)

Haven't done back button yet just for comparison but this look better?

WarrenK-design:
> Hey Warren, looks good to me. should work with the backend no problem.
> 
> only things I noticed are that a '?' pops up when you hover over the days, ie "fri", "thurs" and them. not an issue really, just nothing happens when you click on them. also the start date persists if you go back and change the type of game.
> 
> happy to approve though, i don't think those are actual issues.

I did notice this yeah, its the hover effect on the calendar app, they are formatted as short names but when you hover it shows the long name like "Mon" -> "Monday" I sorta like it. 

Yeah the date one I did have it so that it reset with ```useEffect``` but then like if a user flicks back really quickly I and didn't want the start date/end date to change I taught it would be better to persist than resetting when they click back.  There will eventually be a game summary which shows all these details to them anyway before they complete the game creation.

PR #506, opened 2022-11-14
Backend sell stock route

joeycor92:
Same as buy stock route, only opposite. Assuming ye have bought stock ye can now sell stock using something like this.

{
    "portfolioId": "636d206e2197958e364f05f4",
    "stockId": "63457b65e940127420948c8e",
    "units": 25,
    "buyOrSell": "SELL",
    "orderType": "MARKET"
}

First you can buyStock and then sell it. Try selling a small amount of the stock, then too much of the stock and finally sell the remaining stock.


DaraghK93:
Hi Joey,

I'm getting a weird server error when I try to either empty my holdings of one stock or sell more than I have of that stock
Screenshot 2022-11-15 at 18 25 10
I've tried with a few now. maybe Caolan can try to see if it's not just me. 

Also, for some reason, the remainder of my portfolio doesn't reduce the first time I sell some of the stock that I just bough. below are images of the responses to my buying and then selling the stock straight after. same portfolio remainder. again ye might let me know if I'm just doing something stupid.

Screenshot 2022-11-15 at 18 26 03

Screenshot 2022-11-15 at 18 26 12


Also, change these to the same as in the buy stock one.
Screenshot 2022-11-15 at 18 29 29

One last thing, just on the check if it's market or limit. I thought we were just saying that these were market order routes? should we just keep it to that?

DaraghK93 (changes requested):
just one more thing as well. I don't think transactions updates when a stock is sold. should it?

joeycor92:
> just one more thing as well. I don't think transactions updates when a stock is sold. should it?

I believe that I have covered all of the changes requested.

DaraghK93 (changes requested):
hi Joey,

getting this server error when I trye sell a correct number of stocks (ie ones that I have and have enough of)
Screenshot 2022-11-21 at 12 48 29

not sure what's going on there because the holdings values change alright but nothing else does. The portfolio remainder doesn't change either.

as well, trying to sell more than I have is coming back 200 Ok but then doing nothing. in fact, I didn't get a response form postman from anything I sent. so i think there's something up with the sellStock service. it might not be returning anything. Think you might need to revisit it because I can see a mistake here - 
Screenshot 2022-11-21 at 12 53 22

should be "==" or "===".

Screenshot 2022-11-21 at 12 54 17
I think this should just be its own if, as you have it as an else if from the undefined error check. not sure if that's causing issues but might be more legible.

Screenshot 2022-11-21 at 12 56 21
here as well, we're only sending it as a market order so i don't think there's any need to check for limit. and if you have that if above checking that it's "Sell" and "market" this would be redundant

DaraghK93 (changes requested):
I tried to delete 4 units of a stock I had 4 of and it gave me this 
Screenshot 2022-11-21 at 15 08 53

this happened when I tried to sell more than I have. the error comes up correctly on the console but we're still getting a 200 response. Need to send the error status back to the controller. 
Screenshot 2022-11-21 at 15 12 05

Take a look at leagueServices. you only throw the new error in the catch block. otherwise just return an error message as an object and set the status to that.

Screenshot 2022-11-21 at 15 14 03

then check in the controller
Screenshot 2022-11-21 at 15 16 17

Other than that It seems to be working well

DaraghK93 (approved):
well done Joey, that seems to work very well for me now. great work. 

Just one thing I'll say to make this error a little clearer. This is what it says when you try to sell 6 units of something you only have 5 of. I think the message could be more along the lines of "you do not have that many units to sell". I was just confused as to why it said that when it came up for me.

Screenshot 2022-11-21 at 21 08 47

amplify/backend/function/expressServer/src/controllers/portfolioController.js
```diff
@@ -160,8 +151,113 @@ catch (err) {
...
+      typeof req.body.buyOrSell === 'undefined'||
+      typeof req.body.orderType === 'undefined'
+    ) {
+      // data is missing bad request
+      res.status(400)
+      res.errormessage = 'Details missing for a transaction'
+      return next(
+        new Error(
+          'The client has not sent the required information to buy stock',
+        ),
```
caolan-power, line 171: Only thing I could see was this. buy instead of sell

caolan-power (approved):
All good Joey sorry for the delay here. 

Tested it fully. Even manipulated stock prices to see if all was working and it is working as expected well done

amplify/backend/function/expressServer/src/controllers/portfolioController.js
joeycor92: Will fix this in the next update! Will merge now so that I can move onto the next steps. Thanks.

PR #507, opened 2022-11-15
Machine learning stock recommender system model deployment2

BearachB:
# Changes:
- Not a huge amount of changes file wise since the last PR. 
- Lambda is finally deployed.
- Minor changes to app.py to get this working whilst deployed. 
- Ran into a problem that I couldn't replicate locally. When testing locally, the JSON body file worked as expected, however when testing on the live Lambda function it took the JSON body as a Python Dictionary for some reason. Not sure why this is, it was a frustrating one to fix as locally I couldn't replicate the issue.

# Lambda Test Input "MSFT":
image

# Lambda Test Output:
image

# Next Steps From Here:
- Update stockRoutesService to invoke the recommender service every time a user visits the page. If they have stocks in a portfolio, recommend based on those; if they do not, recommend based on the top 10 biggest positive movers for the day (the specifics here are subject to change). @WarrenK-design you might have the best idea of how to go about this? Does doing it through the stockRoutesService make sense?
- Refine recommendation model using more inputs than just longbusinesssummary.


WarrenK-design (approved):
Hi Bearach good work still works locally. 

On the deployment side of things had a mess around with it there. 

You get the ```forbidden``` or ```Missing Access Token``` response using this URL https://7hkz8cimzd.execute-api.eu-north-1.amazonaws.com/default/ or this url https://7hkz8cimzd.execute-api.eu-north-1.amazonaws.com/default/stockRecommendations (I taught second one would have worked I see you have it in ```template.yaml```). 

The url is actually https://7hkz8cimzd.execute-api.eu-north-1.amazonaws.com/default/stock-recommender-StockRecommenderFunction-Uh3kMlGONr44 I got this from Lambda -> Function -> Configuration -> Triggers 
There are two there as it seems its making any HTTP method go to that function and then the second one is the GET so we may delete the first one later. 

image

It still does get a server error now when testing. 
image

Have a look in cloudwatch logs though the function is running. 

For the request body one I did notice something when trying to test the function in API Gateway. 
image

Instead of sending the data in the body send it as a querystring.

I think too which may solve a lot of preoblems thinking down the line is send a userID as the input. 

GET URL?user=<UserID> 
OR
GET URL Headers user=USERID

Your function uses the user ID to get the users portfolio, makes recommendations on this and then returns these stocks which fit into the stockdiscovery page. May be an advantage too as we can reuse this anywhere now where we have the UserID like calling directly from the user profile page say to show them there recommended stocks. 

Id say too return the full stock objects (minus things like all the prices) because we will need all the data in them to show them. 

```returnFromFunction = [ {stock1, stock2, stock3}]```

Possible flow of program 
1. Parse ```userID``` from query string. 
2. Use ```userID``` to make database call to get stocks in the users portfolio. 
3. Use the stocks returned from step 2 to make predictions. 
     - No stock in portfolio return maybe the top stocks or something? 
5. Get the full stock objects from database using ticker symbol. 
6. Return the full stock objects.

BearachB:
@WarrenK-design Thanks for the review! Yep unfortunately more work needed on it that I hoped for. I was hoping to get it 95% done this week, but deployment has slowed me down massively. Has been very time consuming deploying and debugging along the way.

True if we do it using the user ID as the input it would mean it's a self contained function and just returns the recommendations. I was planning on carrying this all out on the Express side (along with the logic for dealing with a user who owns multiple stocks, or no stocks), but you're right it would be easier to get this all working in one step. I'm working on the third iteration of the model now and will feed these changes in.

caolan-power (approved):
Nice work here nice to see it coming together. 

Was talking to you about deployment etc too. Maybe include a README if anyone else wants to run it too.

PR #509, opened 2022-11-15
Game - Backend - MyGames Route

DaraghK93:
this route -  /api/league/myleagues -  Gets all the user's games and sorts them into :
- Active 
- Scheduled
- Complete

the get function can be found in leagueControllers getMyLeagues. I'm currently sorting by start date descending (newest ones leftmost on the menu) and limiting to 20 leagues. let me know if you think this should change

can test in postman by sending a GET and your token to the above route. if you want to display some games then use the createLeague route and create some that start immediately and some that start in the future. you can create completed games if you'd like by changing the finished field to true in your local database (where active is already false) just to test it out.

I have adapted Warren's frontend to work with the data coming in and it's running with no errors:
Screenshot 2022-11-15 at 17 44 26

My page currently looks like this
Screenshot 2022-11-15 at 17 46 43


WarrenK-design (changes requested):
Hi Daragh, 

Good work tried to break the backend and I couldn't so well done. 

Just on the frontend I think this looks a bit bare. 
image

If you don't mind I have the code to make it look like this, I can send it to you on discord to mess around with or I can commit it to this branch and you can mess around with it, only a small change but looks a little better I think.  
image

Let me know! Well done

DaraghK93:
Thanks for this Warren. I've just edited those slightly. @BearachB have quick look and see what you make of the ui

BearachB (approved):
All looks good @DaraghK93 had a play around on the backend and couldn't fault it.

Good shout @WarrenK-design that's a lot neater and less bare looking. Great to see components being reused, nice easy wins that flesh out the application bit by bit, goes to show how useful it is to make React components properly the first time so they can be re-used again and again.

WarrenK-design (approved):
Looks good Daragh

PR #510, opened 2022-11-15
Tweet_sentiment_visualisation

caolan-power:
Added Warrens query for getting tweets' sentiment to front-end graphs 

+

Linked up front end graphs.

*** Very small PR ***

### DIFF
--- a/amplify/backend/function/expressServer/src/controllers/stockController.js
+++ b/amplify/backend/function/expressServer/src/controllers/stockController.js
@@ -4,6 +4,7 @@ const Stock = require('../models/stock.model');
 /// Imports ///
 const stockService = require('../services/stockRoutesServices')
 const articleService = require('../services/articleService');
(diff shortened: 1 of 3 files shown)


WarrenK-design (changes requested):
Hi Caolan, 

Looks good seems to work fine just the frontend when there are no tweets available. I think it should be more of an info message than an error because it may just be that  no one is tweeting about a company. 

The spacing/height on yours actually looks better so feel free to change the spacing on mine but think it should be a ```variant=info``` and message should be like the new one. To replicate just empty the tweets document locally is probably handiest way. 

image

WarrenK-design (changes requested):
Need to make sure it looks ok on mobile and desktop 

image

WarrenK-design (approved):

beckskelly (approved):
Looks good Caolan, fair play!

PR #512, opened 2022-11-16
Game frontend - Create Game - Page 3 - Balances and Fees

WarrenK-design:
The third page of the create game flow. The sectors page was the 3rd page originally but changed this as I need the portfolio balance before go any further to suggest a winning value for a value based game. 

Reusable component ```RangeSlider``` in widgets folder, going to reuse this on the winning value and ESG ratings screen too so was handy to have. Can be reused in the app too wherever there is a slider later on. 

# Desktop and Mobile Views 
image
image

# Inputs 
The user can use the slider or they can enter a number. They can also use the up/down key on the keyboard to change the number something mentioned in user eval. 
image
If the user tries to enter a number outside of the min/max the state wont be set in the input and an info message will be shown instead. (Awkward to set the state temporarily have seen some use timers online to delay input but think this is fine they will get idea)  
image

# Explanations 
Something that came up in user eval too, they wanted suggestions on what to set for each of these and if they set something what type of game would it result in. 

I broken these down into ```Relaxed```, ```Medium``` and ```Harsh``` with ranges for each. If you click the info button it first gives general overview and then each of the suggestions. 

If we are happy with ```Relaxed```, ```Medium``` and ```Harsh```  I'm going to use it for ESG and Winning Value too just to keep it consistent. (You all know my spelling so just make sure spelling in here ok too, VS code don't have spell check 🤣) 

image
image
image


DaraghK93 (changes requested):
Hi Warren,

I just had a quick look, but there seems to be this overlap in the text and the box when you're inputting the starting balance etc. This only seems to be on the desktop version. mobile was fine for me

Screenshot 2022-11-16 at 13 03 24

Screenshot 2022-11-16 at 13 03 31

Screenshot 2022-11-16 at 13 03 38

DaraghK93 (approved):
all fixed now warren, thanks for those changes. made some slight tweaks to the text in the info buttons

BearachB (approved):
All looks good @WarrenK-design.

One minor thing I spotted is with the boxes for the balances and fees page. Maybe it would look better if the $ sign was inside the box too, instead of being out to the left hand side.? Again, it's a very minor comment and might not be worth spending the time on (also, it's subjective). Up to you!

image

WarrenK-design:
> All looks good @WarrenK-design.
> 
> One minor thing I spotted is with the boxes for the balances and fees page. Maybe it would look better if the $ sign was inside the box too, instead of being out to the left hand side.? Again, it's a very minor comment and might not be worth spending the time on (also, it's subjective). Up to you!
> 
> image

Hi Bearach yeah it would be tricky enough to do because that box is really an ```<input type="number" /> ``` and I've just stripped away a lot of the styling border etc. when not active to just make it look cleaner so inside the box you cant even type non numeric characters. 

I do see your point though but just a tricky one, may go ahead just the way it is for now.

PR #513, opened 2022-11-17
Game Frontend - Create Game - Page 4 - Winning Value

WarrenK-design:
Fourth page of the create game flow.

This page will only appear if a user select a Value Based game, will not appear for duration based games. 
image

# Values
The initial value is based upon the value the user selected for the portfolio balance. It will be a 5% winning value as default. The min a user can go is a 1% increase all the way up to a 15% portfolio increase. 

Example info card for when a user sets a portfolio starting balance of $100,000

image

If the user goes over the 10% portfolio increase in year a warning is shown just telling them it will be a long game. 
image

## Switching between pages
If a user goes back and updates the portfolio balance and the winning value is no longer within these limits (1% to 15%) then it will update. If the winning value is still within the max and min limit (1% to 15%) then it will not change as they may of wanted to keep it. Very annoying clicking between pages and things changing so kept it this way.

### DIFF
--- a/src/components/gameComponents/createGameScreenComponents/GameBalanceFeesTradesSelection.js
+++ b/src/components/gameComponents/createGameScreenComponents/GameBalanceFeesTradesSelection.js
@@ -32,6 +32,7 @@ function GameBalanceFeesTradeSelection({startingBalance, setStartingBalance, tra
                         min={1000}
                         max={1000000}
                         startWidth={"6.75rem"}
+                        showError={true}
                     />
                 </Col>
             </Row>
@@ -60,6 +61,7 @@ function GameBalanceFeesTradeSelection({startingBalance, setStartingBalance, tra
                         min={1}
                         max={300}
                         startWidth={"2.75rem"}
+                        showError={true}
                     />
                 </Col>
             </Row>
@@ -85,6 +87,7 @@ function GameBalanceFeesTradeSelection({startingBalance, setStartingBalance, tra
                         min={1}
                         max={100}
                         startWidth={"2rem"}
+                        showError={true}
                     />
                 </Col>
             </Row>
--- a/src/components/gameComponents/createGameScreenComponents/GameBalanceFeesTradesSelection.js
+++ b/src/components/gameComponents/createGameScreenComponents/GameBalanceFeesTradesSelection.js
@@ -4,10 +4,6 @@ import InfoButtonModal from "../../widgets/InfoButtonModal/InfoButtonModal"
 
 
 function GameBalanceFeesTradeSelection({startingBalance, setStartingBalance, tradingFee, setTradingFee, maxTradesPerDay,setMaxTradesPerDay}){
-    ///
-    //  Max - 1,000,000
-    //  Defautl - 100,000
-    //  Min - 1000
     return(
         <>
             <Card.Title className="gameOptionsCardTitle">Balances and Fees</Card.Title>
--- a/src/components/gameComponents/createGameScreenComponents/GameDurationSelection.js
+++ b/src/components/gameComponents/createGameScreenComponents/GameDurationSelection.js
@@ -18,8 +18,8 @@ function GameDurationSelection({gameStartDate, setGameStartDate, gameEndDate, se
             <Card.Title className="gameOptionsCardTitle">Start Date</Card.Title>
             <Row>
                 <Col>
-                    <Card.Text className="gameDetailsCardText">Games with a start date of today begin immediately</Card.Text>
-                    <Card.Text className="gameDetailsCardText">Scheduled games begin at 9 a.m on the scheduled start date</Card.Text>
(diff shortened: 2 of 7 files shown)


DaraghK93 (commented):
Hey Warren, great work here. really cleverly done. the warning is a great idea and the info cards are great.

I have some comments, just on appearance and usability.

firstly, the way the warning message pops up is a bit unpleasant, in that it shifts the dial down on the page. is there a way to do this where it doesn't alter that component? maybe pops up below it? or just a pop up that they can close once they've read? 

Screenshot 2022-11-17 at 13 24 47

^^ maybe like that would be quickest. 

the other thing is the percentage. i was a bit confused as to what the winning value was defaulted to automatically. could we have the percentage change along with the value change? llke we don't even have to be able to move it, but just the numbers move as the absolute value moves? 

It might also be an idea to explicitly say what the max winning value is at the top. even just in brackets (max value is +10% of starting balance) or something like that.

WarrenK-design:
> Hey Warren, great work here. really cleverly done. the warning is a great idea and the info cards are great.
> 
> I have some comments, just on appearance and usability.
> 
> firstly, the way the warning message pops up is a bit unpleasant, in that it shifts the dial down on the page. is there a way to do this where it doesn't alter that component? maybe pops up below it? or just a pop up that they can close once they've read?
> 
> Screenshot 2022-11-17 at 13 24 47
> 
> ^^ maybe like that would be quickest.
> 
> the other thing is the percentage. i was a bit confused as to what the winning value was defaulted to automatically. could we have the percentage change along with the value change? llke we don't even have to be able to move it, but just the numbers move as the absolute value moves?
> 
> It might also be an idea to explicitly say what the max winning value is at the top. even just in brackets (max value is +10% of starting balance) or something like that.

Hi Daragh have updated it there according to changes, as mentioned couldn't get it outside the card due to the internal state but think it works. 

image

I haven't explicitly said the max value in the card as it looked a bit cluttered but if they move slider all the way to the end it will show them the max in the sentence, the warning on the slider still pops up too if they try to enter anything under/over max/min as on the previous pages. 

Let me know what you think!

DaraghK93 (approved):
Hey Warren,

I think that looks a lot better. the warning popping up underneath looks a lot less jarring. and the percentages are very clear.

Just one thing i noticed on the form for entering the amount. you can't actually enter the amount yourself, as you can't add or delete a number from it, as doing so would mean the range was greater than the 1 %- 15% range. so essentially you can still only use the slider or the arrows on that form. I went back and checked the other sliders too, they're the same. so as long as there's a check for max and min, you won't be able to edit it yourself by adding or deleting figures in the form. Let me know if you have questions on this. I'm happy to approve for now as this might be a component wide thing that can be fixed later.

BearachB (approved):
Looks good @WarrenK-design, I like it. Technically it all works and have no problems running it etc. Although I do have a few comments on the overall usability of it. 

# Not being able to enter number manually
I know for myself personally, I'd never want to enter a number by using a slider and would always opt for typing in the number (but I might be an outlier there).
You can't actually enter the dollar amount in the box here and either have to use the slider or use the up and down arrow (not sure of the specific name) to increase/decrease the amount $1 at a time, which stood out to me anyway. I know we're starting to get stuck for time so we shouldn't be wasting time fine tuning everything as we go, it works so I'm happy to approve but I think if we have time this should be revisited to allow for selection either by typing in the value or using the slider. This applies to the previous "Balances and Fees" screen too.

# Screenshots
## When trying to type in a value
image

WarrenK-design:
Ok I think I fixed it. You can type in values now and the ```next``` button will be disabled if there is an error.  

# Balances and Fees 
image

# Game Winning Value 
For the game winning value gave more detailed error to explain where the min and max are coming from. It highlights the numbers which need to change in red too. 
image
image

If you go back and adjust the starting balance to a value which makes the winning game balance out of range, you wont be able to advance. 
To get what I mean do the folllowing steps. 
1. Set Portfolio Balance to $100,000
2. Set the Winning Value to $110,000
3. Now go back and change the Portfolio balance to $1,000,000
4. Winning balance no longer in range you should not be able to advance until you change it 

Let me know if spot anything else, cheers!

DaraghK93:
Hi Warren, Thanks for those changes. I think that's a good work around, well done

BearachB:
All good @WarrenK-design thanks for the update!

Happy to approve, but again one or two comments on the broader strokes.

# Game Winning Value - Only a max winning value of 15% ROI seems quite low to me. 
image
Why not give them the option to go higher? It's not unheard of for companies (especially tech. stocks such as NVDA or AMD) to have a double digit % gain/loss in a day, which might lead to some very short games (not necessarily a bad thing, but why not give them an option). Even just looking at our "Today's Biggest Positive Movers" section, Disney is up 10% since yesterday after they fired their CEO and appointed a new one.
 
image

Again, not a priority now and we need to not get bogged down in the tiny details right now, but it might be worth adjusting this down the line.

PR #517, opened 2022-11-22
Game Frontend - Create Game Page 5 -  Stock Type Screen

WarrenK-design:
Smaller PR than last one, this is for selecting the sectors for the stocks. 

I didn't think it needed an info modal bit less complicated than the rest of the pages for setting balances/fees but if it needs more explaining let me know. 

# Desktop 
image

# Mobile
image

# Error 
image


DaraghK93 (changes requested):
Hey Warren, 
Looks good. I'm unsure about having all sectors preselected, but i don't think it's an issue really. it's probably easier to do it that way as it pushes them to select more, in which case more stocks are available to them. 

the only real change i think you should make is to explain it more than "select tradable stocks". I think they should be told something like "select the sectors players can trade with in the league. by not selecting a sector, you disallow players from trading companies in that sector" 

That mightn't be the best way to word it but i do think it warrants slightly more explanation

WarrenK-design:
> Hey Warren, Looks good. I'm unsure about having all sectors preselected, but i don't think it's an issue really. it's probably easier to do it that way as it pushes them to select more, in which case more stocks are available to them.
> 
> the only real change i think you should make is to explain it more than "select tradable stocks". I think they should be told something like "select the sectors players can trade with in the league. by not selecting a sector, you disallow players from trading companies in that sector"
> 
> That mightn't be the best way to word it but i do think it warrants slightly more explanation

Cheers Daragh, 

Updated that there. 

image

DaraghK93 (approved):
looks good Warren, thanks for the change

BearachB (changes requested):
@WarrenK-design Looks good! Great progress on this so far.

Just two UI aspects that should be easy fixes and make the page a lot more usable. 

# Could you make all the selection buttons the same width and height. 
It's pretty confusing at the minute and the fact they're all different looks quite messy. Could you make them all  the same width and height? Also might be worth centring the text inside them. See my crude mockup below:
image

# Could you change the colour of the selection buttons so that there is a difference between he industry selections and the "Back" and "Next" buttons. 
At the moment they're exactly the same and it would be easy to accidentally click on the nav buttons instead of the industry buttons. Again, see crude mockup below (I only coloured the bottom row of them to get across the idea):
image

BearachB:
> > @WarrenK-design Looks good! Great progress on this so far.
> > Just two UI aspects that should be easy fixes and make the page a lot more usable.
> > # Could you make all the selection buttons the same width and height.
> > It's pretty confusing at the minute and the fact they're all different looks quite messy. Could you make them all the same width and height? Also might be worth centring the text inside them. See my crude mockup below: image
> > # Could you change the colour of the selection buttons so that there is a difference between the industry selections and the "Back" and "Next" buttons.
> > At the moment they're exactly the same and it would be easy to accidentally click on the nav buttons instead of the industry buttons. Again, see crude mockup below (I only coloured the bottom row of them to get across the idea): image
> 
> Hey Bearach,
> 
> Good suggestions this ok?
> 
> image

Yep that looks much neater I think!

BearachB (approved):

PR #518, opened 2022-11-22
Backend buy stock fees

joeycor92:
Well folks. Hopefully not a bad one.

So we now have the fees included where it is taken off the remainder each time. For the buy stock it also has to check that this won't go into the negatives.

As well as this, there are a number of other checks:
ESG ratings, checks if this is below the min rating provided when creating the league.
MaxTrades, checks if this hasn't been exceeded

The number of trades completed is also incremented each time a trade happens using $inc.

To create a portfolio for this, use the createLeague route:
http://localhost:3001/api/league/createleague

Sample input
{
    "leagueName": "TestForFees",
    "startingBalance": 10000,
    "leagueType": "timeBased",
    "tradingFee": 1,
    "private": true,
    "startDate": "2022-11-23",
    "endDate": "2022-11-30", 
    "maxDailyTrades": 3,
    "image": "/stock_photo_1.jpg"
    }

Example of creating a league with specific sectors:
{
    "leagueName": "TestForFees",
    "startingBalance": 10000,
    "leagueType": "timeBased",
    "tradingFee": 1,
    "private": true,
    "startDate": "2022-11-23",
    "endDate": "2022-11-30", 
    "maxDailyTrades": 5,
    "image": "/stock_photo_1.jpg",
    "sectors": ["Basic Materials", "Communication Services"]}

You'll get a portfolio from this which you can then use in this. The stockId is Google so that will work with the communication services sector from above. But you can use any others to make it break.
{
    "portfolioId": "637d3463d3f71d738e13a730",
    "stockId": "63457b65e940127420948c9f",
    "units": 1,
    "buyOrSell": "BUY",
    "orderType": "MARKET"
}

Let me know if you have any issues.

Next on the list is limit orders.


caolan-power (approved):
Hi Joey.

Really good work I tried breaking it and wasn't able.

The only thing I'll flag and this is probably more for the game as a whole. 

I was able to create a league where the trading fee was greater than the starting balance we should look into that.

But your code caught it as normal and returned an insufficient funds errorrmessage which is perfect

DaraghK93:
Hey Joey,

Great work. couldn't break it at all.

@caolan-power mentioned the max trading fee. that was in my controller so I can fix that.

PR #519, opened 2022-11-22
User settings page

caolan-power:
This adds the functionality for a user to change their details. It also includes some feedback changes for other parts in the app. Small things like making it easier for the user to know what the password requirements are and also a 'show password' check box. 

the user settings page can be accessed right beside the logout button and to make any changes the current password must be included

here are some examples of the UI. Note i'm using the images from the game creation screen for now as I haven't found any nice ones yet. However, this will be a matter of adding them to the public folder and the strings will be updated as normal.
Screenshot 2022-11-22 at 22 00 03

Screenshot 2022-11-22 at 21 59 50
Screenshot 2022-11-22 at 21 59 54

I've tried adding useful error message alerts where possible and useful errors returning from the backend. I think i've tried most ways of breaking it but it's time now to get others looking at it and let you try break it. I'm sure there's something i've missed.
Screenshot 2022-11-22 at 22 05 26

Sample message as shown above. Open to changes on this but I included the something gone wrong one at the bottom as if you're at the bottom of the page sometimes a user wouldn't know something is wrong then they can scroll up to see what it is specifically.


beckskelly:
Put it into a card for a border if you want it, like the following:

<Card>
<Container>


<--- your stuff---->

<Container />
<Card />

beckskelly (changes requested):
Also getting the same compile error as @BearachB. I agree with @Bearach in that profile screen would be nice as well, seeing as the profile link on the nav goes nowhere, but this can be handled in a separate PR

beckskelly:
Also, why can't I change my name out of interest?

BearachB (approved):

BearachB:
That's great @caolan-power thanks for those changes. I think they'll make the user's life a bit easier.
As Discussed on Discord, I've pushed a very small update to some of the text on the page to make it a bit clearer to the user. If you could tag me for re-review when you update the redux bits to update the current account information on change.

WarrenK-design:
Just too @caolan-power have a look at this website https://www.svgrepo.com/ for the game avatars, think might look bit bland having avatar images and game images the same.

caolan-power:
Ok I have added changes for updating redux. Added new images for the avatars. Added user details showing. Added the first and second name fields. Think I've addressed all comments. If I forgot anything or you spot a bug let me know. Going to tag Warren too to check the redux stuff out

On the redux stuff I ran into an error when copying the token code in login or register controller so I'm just sending the same token back as the one received so hopefully that is ok to do. Seems to work for me now

caolan-power:
Screenshot 2022-11-24 at 16 08 16
Screenshot 2022-11-24 at 16 08 23
Screenshot 2022-11-24 at 16 08 30

caolan-power:
Check all error messages too and check if they're working the same. Also me and Warren talked about email case sensitivity during the week so will add that in later. Will add the new password fields to the other forms where it makes sense too.

beckskelly:
Nearly perfect @caolan-power just one small cosmetic issue, id get rid of these underlines here and just have the fields as bold,  the rest of the app has nothing underlined 
image

beckskelly (approved):

WarrenK-design (changes requested):
Hey Caolan, 

Good work coming together well, few things. 

# Errors
Maybe I'm doing something wrong but tried to change some of my details and I get this error. 

It doesnt tell me what is wrong either which is bit worrying, should always give user some indication of what they have done wrong if we can. 

image
image

In the backend this is the error that is getting printed. 
image

For this too you set error messages on the back but then it seems no matter what the error is it is printing out "Something Went Wrong". Should be thing like "Password Incorrect" "Password.
does not meet the required lenght" something to give user indication of what went wrong. 
image.

# Email
You show the email up the top but there is nowhere I can change this in the form, Id say its one of the more important ones too if a user updates there email address should be able to change it.

PR #520, opened 2022-11-23
Machine learning stock recommender system model model refinement

BearachB:
There has been 2 main changes to this:
- Updates to app.py:
  - These updates were made to allow the deployed version to work. There was a problem with the JSON datatypes being handled differently on the dpeloyed version for some reason (could not figure out why). This was solved by converting to JSON using json.loads(). I've left in some print functions for now as they can help with debugging from within Cloudwatch.
  - This live API now functions.
    - GET request to API endpoint - "https://7hkz8cimzd.execute-api.eu-north-1.amazonaws.com/default/stock-recommender-StockRecommenderFunction-Uh3kMlGONr44"
    - Request body: `{"stock": "MSFT"}`
    - image
    - No authentication needed yet, this will also be added.

- Updates to model_3.py
  - get_user_stock function has been added to the model, this means that the input will now be a username and from there the API will fetch their stocks using PyMongo and then run the recommender system on those stocks.
  - More inputs have been added to the model bag of words; sectors, industries and countries and exchanges. This was done as it was quick to add them in. The other features such as ESG ratings should be added in today. I want to first get this all working with the backend/frontend first and then make changes.
  - You will not be able to run this without mocked data, if you'd like to get onto me and I can give you the JSON dumps for this. However I don't think that necessary as this code is not deployed yet so isn't running.
  - This model_3.py has not been added to app.py (i.e. the deploted version) yet.


machineLearning/stockRecommendationSystem/modelGeneration/model_3/model_3.py
```diff
@@ -0,0 +1,216 @@
...
+    # Read in the s&p dataset
+    stocks = pd.read_csv('machineLearning\stockRecommendationSystem\modelGeneration\model_3\stocks.csv')
+
+    # Remove some columns that won't be used
+    stocks.drop(['idnumber','longnamesort', 'weight'], inplace=True, axis=1)
+    # print(stocks)
+    print(stocks.info())
+    # Create array X of the all long business summaries, sectors, industries and countries and exchange
+    X = np.array(stocks.longbusinesssummary + " " + stocks.sector + " " + stocks.industry + " " + stocks.country + " " + stocks.exchange)
+
```
caolan-power, line 23: Just a thought is the exchange a meaningful input? the snp only has stocks from like 3 exchanges. Also country we only have 19 not from America. Wonder would it cause any issues with overfitting or never include similar stocks to non american stocks if that's the case. Could be wrong just food for thought anyway.

caolan-power (approved):
Seems to work as expected. Fair play. Next deployment of getting user details in will be interesting.

Also I never realised username was a unique field so i'll need to go back to my own work and account for that

WarrenK-design (changes requested):
Hi Bearach, 

Works fine on live version but I cant actually run it locally, just be nervous if every time you make a change we have to deploy it to test it, how are you testing it locally? 

image

machineLearning/stockRecommendationSystem/modelGeneration/model_3/model_3.py
```diff
@@ -0,0 +1,216 @@
...
+        # Setting the DB name
+        db = mongoClient.dev
+        # Setting the DB collection names
+        user_collection = db['users']
+        portfolio_collection = db['portfolios']
+        holdings_collection = db['holdings']
+        stocks_collection = db['stocks']
+
+        # Get user information
+        user_data = user_collection.find_one({"username":userName})
```
WarrenK-design, line 88: Id recommend using the user ID here, will improve the query speed Id say

machineLearning/stockRecommendationSystem/modelGeneration/model_3/model_3.py
BearachB: It actually doesn't change the output at all because those inputs get diluted down when compared to the longbusinesssummary. But yep can remove them either, my thinking was that we have the data there and it's easy to add to the "bag of words" that are being compared so might as well.

BearachB:
> Works fine on live version but I cant actually run it locally, just be nervous if every time you make a change we have to deploy it to test it, how are you testing it locally?

The live version that is currently deployed I didn't actually test locally, it's because the JSON input seems to be handled differently when running locally vs deployed (I don't understand why this is as it should be exactly the same environment). This was the main roadblock for getting this deployed and what has slowed me down the most as it would be working locally but then when deployed would throw an error. 

What I did in the end was add debugging print statements to the deployment, then deploy and test on the live API. This really was not ideal, however it was the only way I could get the live API to work.

WarrenK-design:
Ok so as discussed on Discord @BearachB, 

With that last commit I added if ```ENVIRONMENT``` is ```prod``` do all your production stuff in that if loop, else if ```ENVIRONMENT```` is ```dev``` read just the local config stuff.

When running the to overwrite the environment cars use ```sam local invoke --env-vars env.json```, the ```env.json``` file is just in format of this [here](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-using-invoke.html#serverless-sam-cli-using-invoke-environment-file)

If you want to use the custom event file can use ```sam local invoke --env-vars env.json --event .\events\eventMSFT.json```

Let me know if it works!

BearachB:
> Let me know if it works!

Awesome thank you @WarrenK-design that simplifies it a bit! All working locally!

PR #525, opened 2022-11-23
Backend - Game - updateCreateLeague

DaraghK93:
real small one here. I've just changed the check for starting balance and trading fee. 

trading fee must now be between 0 and 300
starting balance must be between 1000 and 1000000

Screenshot 2022-11-23 at 12 13 07

Screenshot 2022-11-23 at 12 12 48


caolan-power:
Nice one any reason you chose those numbers? Are they arbitrary? Or should max fee be a percentage of the starting balance? or is this to accomodate front end?

caolan-power (approved):
LG2M R2M

joeycor92 (approved):

PR #526, opened 2022-11-23
Database code for creating views

joeycor92:
Similar to Daragh's PR there. This is the code needed to have the views locally. Everyone will need to implement these.

Should be done through 'Aggregation' found in MongoDB Compass.
image

You choose one of the headings from the drop down menu e.g. $lookup
image

And then create a view at the end.
image

This will produce the views that can be seen on the main database right now.


DaraghK93 (approved):
Good work Joey, they've been working very well for my queries

PR #527, opened 2022-11-23
Game Frontend - Create Game Page 6 - ESG Restrictions Screen

WarrenK-design:
Sixth page in the create game flow for ESG restrictions. 

No sliders for this page as discussed issues could arise where game has no stocks to trade/not a great game. 

4 selections with mins for each
1. No restrictions
2. Environment
3. Social
4. Governance

# Desktop View
image

# Mobile View 
image

# Warnings 
You will get these warnings in the react console for unused state, I am only setting them in this PR, I will use them in the next PR for sending the POST request but need to initialise them in the PR. 

image


BearachB (approved):
Really good work @WarrenK-design. This is exactly the direction we need to go in to make it feel more like a game!

Can't spot anything out of place here at all. I might have missed it earlier, but what made you set the E, S and G thresholds as you have? Like why set environment to 4/5, social to 2.5/5 and governance to 3.5/5?

WarrenK-design:
> Really good work @WarrenK-design. This is exactly the direction we need to go in to make it feel more like a game!
> 
> Can't spot anything out of place here at all. I might have missed it earlier, but what made you set the E, S and G thresholds as you have? Like why set environment to 4/5, social to 2.5/5 and governance to 3.5/5?

Hi Bearch, 

Sorry I missed this, it was based on this table Daragh made, had to find balance between having enough stocks but still making it restrictive.  

image

PR #530, opened 2022-11-23
Individual stock page - changes as per User evaluation and expert reviews

beckskelly:
Fixed many small issues on the individual stock page, as follows:

1. Deleted follow button
2. Changed bottom button text to "Go to trade page!"
3. Made links clear for esg component
4. Made links clear for news component
5. Added information button and links for sentiment analysis
6. Made buttons stay active on pricegraph
7. Made all prices on the page 2 decimals

Its all very simple stuff, the main thing I want reviewed / opinions on is if the information button on twitter and news sentiment is ok? So far, they are the same and just explain what sentiment analysis is with a link. Should this be more personal or gamey?

image
image
image
image


DaraghK93 (changes requested):
Hey Becks, great work. it's great to tick a few of these off.  a few comments just.

2. I think we discussed "Let's Trade!" on friday. I think it's a little more gamey and snappier too. also I think people mightn't know what 'trade page' is. Open to other suggestions but i think shorter the better.

5.  Thanks for adding the info buttons, 100% necessary. I think that the actual info might need to be less computer sciencey, so that most people wiill actually understand it. If we could find one that talks about it a lot less formally and/or talks about it in the context of trading, it would be better. I'd even put together a few lines to tie some theory. Something like: "See how this company is doing in the media! We've analysed recent news and tweets to how each company is faring in the media. To learn more about how sentiment analysis works, click here!" 

Well done on everything else. it looks very well. One thing I did see that the one week price graph isn't returning anything. not sure if that had anything to do with your decimal change or anything. probably not.

WarrenK-design (changes requested):
Hi Becks,

Good to get started on these, just few things. 

# Sentiment Explanation 
Yeah echo Daraghs comment no words like "NLP" or "Machine Learning". 

Could just be a sentence at top, "This information relates to how the companies Sentiment in news Outlets", then for twitter  "This information relates to how the companies Sentiment on the website Twitter." Maybe one more sentence then "This information can aid you in making decisions on stocks or something"  just to explain why it is there. 

Break down what Positive, Neutral and Negative means then sorta like this "Relaxed", "Medium" "Harsh" but maybe use the colours the badges use like Red-> Negative, Orange->Neutral, Green->Positive 
 
image


Then at the very end "If you would like to learn more about Sentiment Analysis click here"

beckskelly:
Hey @DaraghK93 and @WarrenK-design, thank you for the reviews! I changed the bottom button to "Lets Trade!", its better, and the sentiment information as well. Let me know if this is what you guys are after:

image

WarrenK-design (approved):
Hi Becks, 

Looks good I just changed small things. 

For text formatting too it may save you in future instead of using lots of ```<br>``` tags, wrap text in a ```<p>``` tags and it will format with padding, if need more then use CSS. 

If we get time/have to reuse that ```CharCard.js``` anywhere in the app in future I think we should pass the information in ```infoModal``` as a prop, at the moment we can only use the infoModal with it if the Title is "News Sentiment" or "Twitter Sentiment", not a big deal but just makes it less reusable. 

Good Work!

DaraghK93 (approved):
Well done Becks, looks great. Thanks for making all those changes

PR #533, opened 2022-11-23
Backend - Game - trades and valueHistory Triggers

DaraghK93:
I've added two more triggers to the Cluster. 

### Reset Trades
This trigger goes off at midnight every night and resets the todaysTrades field on a portfolio. this is for checking whether they've reached their maxDailyTrades in that league.  seen below

Screenshot 2022-11-23 at 21 59 38

### Push Value to Portfolio History
this trigger also goes off every night just before 12 and pushes the portfolio value and today's date to valueHistory array field. this will allow us to create the graphs on the portfolio page. This is a long query so I won't post it here.

They're currently working on the live database. It will be easier to test them further on the live site now that we have the create game frontend done


joeycor92 (approved):
Happy out with this. The logic makes a lot of sense to me and as you said it's already working on the database. Thanks for this.

BearachB (approved):
Good work on this @DaraghK93, already deployed and working on the database so good to see progress being made!

PR #534, opened 2022-11-24
Game frontend - Create Game Page 7- Game Summary Page

WarrenK-design:
Last page in the create game flow shows the summary and creates the game. 

# Desktop view 
image

# Mobile View
image

# Creating a Game
When you click the "Finish" button it will create the game if there are no errors returned.  For the moment it just redirects you back to the "Game" page and the game should be there in the "Active" or "Inactive" section. Will eventually redirect you to the individual game page instead I think that will be better flow. 

For the moment too the game returned isnt stored in any state but I just have the result printing to the console if you want to view it. 

image

image

### DIFF
--- a/src/components/gameComponents/createGameScreenComponents/GameCreationOptionsCard.js
+++ b/src/components/gameComponents/createGameScreenComponents/GameCreationOptionsCard.js
@@ -6,10 +6,12 @@ import { APIName } from '../../../constants/APIConstants'
 import { useState } from 'react';
 import { API } from "aws-amplify";
 import {useSelector} from 'react-redux';
+import MessageAlert from "../../widgets/MessageAlert/MessageAlert";
+import LoadingSpinner from "../../widgets/LoadingSpinner/LoadingSpinner";
 
 
 /// GameCreationOptionsCard ///
-//  prprs:
+//  props:
 //      children         - The children of the component https://reactjs.org/docs/composition-vs-inheritance.html
 //      setScreen        - Setter to set createGameScreen state
 //      screen           - The screen state set by setScreen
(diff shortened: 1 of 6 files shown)


BearachB (approved):
That's really good @WarrenK-design, very nice to be able to create a game!
Can't spot anything out place on it that needs to change. I know we had discussed this last Friday but just wondering if we're leaving the league naming as you have it on the "Basic Details" page with the dropdown?
image

I know letting a user enter a name could open it up to the possibility of inappropriate names, but I think it'll make the names much easier to follow.

WarrenK-design:
> That's really good @WarrenK-design, very nice to be able to create a game! Can't spot anything out place on it that needs to change. I know we had discussed this last Friday but just wondering if we're leaving the league naming as you have it on the "Basic Details" page with the dropdown? image
> 
> I know letting a user enter a name could open it up to the possibility of inappropriate names, but I think it'll make the names much easier to follow.

Yeah I really don't mind know at this stage, when I created like 5 games it started to get hard to follow alright

joeycor92 (approved):
Very good. One thing I might suggest for future work is that this should probably go up in thousands or even hundreds instead of ones, but I understand that would probably cause issues with the input and the slider so not that big a problem. I actually wasn't fully aware that you could input the amount either, I just so happened to click on it and then noticed.
image

The other thing that I've discussed with you is the Sectors, they should probably be unselected at the start as the tab is called 'Select Sectors' so you would intuitively start selecting the sectors you want to include. 
image


These are both small things and doesn't affect the fact that it looks brilliant and it works. Thanks very much for everything on this.

PR #536, opened 2022-11-24
adding css for tweet card

caolan-power:
Tiny PR to add the CSS back to tweet cards after I had globally set all tweets to have no styling.

Sorry about that


joeycor92 (approved):
It works, only thing I'd note is that it's not consistent with the SCSS throughout which uses .componentName instead of #componentName if that makes sense. Only a small thing really.

WarrenK-design (approved):
Hi Caolan, 

Yeah looks good, only thing yeah as Joey said we have been using classes instead of id's for CSS but not a big deal.

PR #538, opened 2022-11-24
Backend limit order routes

joeycor92:
Another update to the Buy and Sell stock routes.

This time it introduces the idea of limit orders. They follow these rules we set out before in the Discord:
For Limit Buy Order:
Funds come out of remainder
No Holding Created
Transaction says pending
When limit reached, transaction says complete, holding created

For Limit Sell:
Remainder stays the same
Unit holdings decreases
Transaction says pending
When limit reached, transaction says complete, remainder increases by the amount of the sale

For testing it probably makes sense to test buy market, sell limit, sell market, and then you can try buy limit whenever.

Routes and examples:
Create League - http://localhost:3001/api/league/createleague
{
    "leagueName": "TestForFees",
    "startingBalance": 10000,
    "leagueType": "timeBased",
    "tradingFee": 1,
    "private": true,
    "startDate": "2022-11-25",
    "endDate": "2022-11-30", 
    "maxDailyTrades": 20,
    "image": "/stock_photo_1.jpg"
    }

Buy a stock (Market Order) - http://localhost:3001/api/portfolio/buyStock
{
    "portfolioId": from create league,
    "stockId": "63457b65e940127420948c9f",
    "units": 10,
    "buyOrSell": "BUY",
    "orderType": "MARKET"
}

Buy a stock (Limit Order) - http://localhost:3001/api/portfolio/buyStock
{
    "portfolioId": from create league,
    "stockId": "63457b65e940127420948c9f",
    "units": 10,
    "buyOrSell": "BUY",
    "orderType": "LIMIT",
    "limitValue": 90
}

Sell a stock (Market Order) - http://localhost:3001/api/portfolio/sellStock
{
    "portfolioId": "637fa445f6ca5fa9a25df90a",
    "stockId": "63457b65e940127420948c9f",
    "units": 9,
    "buyOrSell": "SELL",
    "orderType": "MARKET"
}

Sell a stock (Limit Order) - http://localhost:3001/api/portfolio/sellStock
{
    "portfolioId": "637fa445f6ca5fa9a25df90a",
    "stockId": "63457b65e940127420948c9f",
    "units": 9,
    "buyOrSell": "SELL",
    "orderType": "LIMIT",
    "limitValue": 100
}


DaraghK93 (changes requested):
Hi Joey, 

nice work, I think it's nearly good to go. I didn't get anything very unexpected, I think you've accounted for most things as far as i can see. I'l continue to test edge cases anyway when I think of them.

one thing  I noticed is that when you  make a limit order the value of the transaction is incorrect. It's just the value of one unit, as opposed to units*value, see below. In market orders they're correct so should be an easy fix i think.

Screenshot 2022-11-27 at 13 59 50

Another thing I noticed is that if you make an order to sell a stock by limit order, the overall value of your portfolio goes down by the value you've sold. I don't think that's ideal, because realistically they potentially have more money (bar the tradingfee), but it's not reflected in the total value until the remainder goes up when the limit sell is actioned. 

I think the way around this is adding to the holdings schema maybe. splitting it up like available holdings and a pending holdings or something. where the total value of that holdings is the sum of them, and the pending holdings will go down to 0 once they're sold. so when they're sold, remainder goes back up.

anyway we can discuss this tomorrow maybe

joeycor92:
> Hi Joey,
> 
> nice work, I think it's nearly good to go. I didn't get anything very unexpected, I think you've accounted for most things as far as i can see. I'l continue to test edge cases anyway when I think of them.
> 
> one thing I noticed is that when you make a limit order the value of the transaction is incorrect. It's just the value of one unit, as opposed to units*value, see below. In market orders they're correct so should be an easy fix i think.
> 
> Screenshot 2022-11-27 at 13 59 50
> 
> Another thing I noticed is that if you make an order to sell a stock by limit order, the overall value of your portfolio goes down by the value you've sold. I don't think that's ideal, because realistically they potentially have more money (bar the tradingfee), but it's not reflected in the total value until the remainder goes up when the limit sell is actioned.
> 
> I think the way around this is adding to the holdings schema maybe. splitting it up like available holdings and a pending holdings or something. where the total value of that holdings is the sum of them, and the pending holdings will go down to 0 once they're sold. so when they're sold, remainder goes back up.
> 
> anyway we can discuss this tomorrow maybe

Thanks Daragh, good spots. Think I've covered these now. 

I added in frozenHoldingsUnits to the holdings model. There will need to be a change to the views we had created to account for this addition. Let me know what you think.

amplify/backend/function/expressServer/src/services/portfolioServices.js
```diff
@@ -71,32 +73,48 @@ const buyStock = async (buyData, portfolioRemainder,value, transactionFee) => {
...
         // create transaction object
         await transaction.save()
+        let newRemainder
+        if (transaction.orderType === "MARKET"){
+            // update the db for market order
+            newHoldings = await Holdings.updateOne({_id: holdings._id}, {units: currentHoldings})
+            newRemainder = portfolioRemainder - transaction.value - transaction.tradingFee
+        }
+        else {
+            newRemainder = portfolioRemainder - transaction.value*transaction.units - transaction.tradingFee
```
caolan-power, line 89: Should this be multiplied? Think this is my problem?

DaraghK93 (changes requested):
Hey Joey, I think it's very nearly there now. Just another thing I noticed, it's more or less the same comment as the limit sell one but the opposite. 

When the user submits a limit buy order they're remainder goes down to signal that that cash is now frozen. However because of this the totalValue also goes down. I think a frozenCash field is necessary so that the total value doesn't decrease (except by the amount of the trading fee). 

so the totalValue will be spendableMoney + frozenMoney + holdingsValue

caolan-power (changes requested):
Hi Joey I'm still getting a similar error where the remainder sky rockets down on limit buys

I've noticed now it's fine when doing just limit buys but when I put in a market buy then a limit buy it happens

joeycor92:
Ok I've made a few changes there that I hope should cover everything. I had to make it so that frozenUnitHoldings and frozenBalance are both created from the start. I had mentioned that I need to check whether a user has already had a buy limit order in place for a specific stock but that isn't the case. 

Hopefully all ok now.

caolan-power (approved):
HI Joey did a good bit of testing here and I think you've cracked it finally. Well done happy to approve now

DaraghK93 (approved):
Hi Joey, Well done here. that was a long one. I think I've tested it as much as i can and haven't found anything. Think we're ready to start getting this working on the frontend, so that's one step closer. 

Great work!

PR #540, opened 2022-11-24
Backend - portfolio - get user's portfolio 

DaraghK93:
This is the route and controller for getting the user's game portfolio. 

This will be used in a number of places. two of which are:
- on the game screen, to get the portfolio for the user to view
- on the trade page, so the user can see what they have in that portfolio - balance, holdings, how much of a stock they can sell etc.

The route is portfolio/:userId/:leagueId

we will have both of these in redux, or at least the user Id, and the gameId will be selected when they click onto the game, or click the dropdown with that game on the trade page.


amplify/backend/function/expressServer/src/controllers/portfolioController.js
```diff
@@ -293,10 +293,130 @@ catch (err) {
...
+      res.errormessage = 'No portfolio found'
+      return next(
+        new Error(
+          'No portfolio found'
+        )
+      )
+    }
+
+    //check if the token is their token and corresponds to the params
+    if(req.user.id !== req.params.userId){
```
WarrenK-design, line 313: This check isn't needed if we get the userID from the auth middleware instead I think. The request will never get this far if unauthorised and middleware will catch this. If we remove userID then too from route like in suggestion wont have to check req.params.userId either.

WarrenK-design (changes requested):
Hi Daragh, 

Just one small change and sorry should of spotted iton call we had earlier, we never store the user ID on the frontend, we store the JWT. 

Could you change this to ```GET /api/portoflio/:portfolioID``` and then as it is a protected route get the UserID from the JWT like the other routes?

DaraghK93:
Thanks @WarrenK-design, that's fixed now

DaraghK93:
@joeycor92  when you're looking at this, the route has changed to api/portfolio/:leagueId

WarrenK-design:
Hi Daragh, 

Think I'm running into problem Joey was. 

So Ive created a league. 
```json
{"_id":{"$oid":"63848e126d26665cc1d32fe3"},"leagueName":"The Big Stock League","startingBalance":1000,"leagueType":"valueBased","private":true,"active":false,"finished":false,"startDate":{"$date":"2022-12-31T00:00:00.000Z"},"winningValue":1001,"sectors":["Basic Materials","Communication Services","Consumer Cyclical","Consumer Defensive"],"maxDailyTrades":10,"tradingFee":10,"minERating":0,"minSRating":0,"minGRating":0,"accessCode":"hpzvk9","image":"/stock_photo_1.jpg","leagueAdmin":{"$oid":"63848d4b6d26665cc1d32fd8"},"users":[{"$oid":"63848d4b6d26665cc1d32fd8"}],"portfolios":[{"$oid":"63848e126d26665cc1d32fe5"}],"__v":0}
```
This did make a portfolio then too. 
```json
{"_id":{"$oid":"63848e126d26665cc1d32fe5"},"portfolioName":"The Big Stock League Portfolio","remainder":1000,"startingBalance":1000,"holdings":[],"dateCreated":{"$date":"2022-11-28T10:31:46.473Z"},"transactions":[],"leagueId":{"$oid":"63848e126d26665cc1d32fe3"},"userId":{"$oid":"63848d4b6d26665cc1d32fd8"},"tradesToday":0,"valueHistory":[],"__v":0}
```

Im sending the request ``` GET /api/portfolio/:leagueId``` with auth token in the header but it returns ```No portfolio found```.
image

The controller does get the league ID and the User ID but the mongo query seems to return ```[ ]```

image

Its this code which is being hit. 

image

I ran the query too just directly on Mongo and seems to be returning ```[ ]``` too. 

image

WarrenK-design (approved):
Looks good Daragh, all working now after adding views locally.

PR #541, opened 2022-11-24
removed pre save as double hashing passwords

caolan-power:


WarrenK-design (approved):
Think has fixed it!

DaraghK93 (approved):
seems to have fixed. thanks for that!

PR #542, opened 2022-11-24
rehashing password for password recovery

caolan-power:


WarrenK-design (approved):
Tested Register -> Logout -> Change Password -> Login and all seems to be ok

DaraghK93 (approved):
all works perfectly! thanks for fixing

PR #546, opened 2022-11-25
Stock discovery screen revamp

beckskelly:
Its the weekend so theres no need to review this until Monday.

Changed up the stock discovery screen as per Andreas comments. I basically widened the containers to fit the entire screen with padding on either end.  I tried changing the background colour but I felt it looked a bit dirty, so I left it as is for now, but can always change it to a background color with no border on the cards if you think its bettter

Let me know what you think.
image
image

Attempt with background color changed:
image

### DIFF
--- a/src/components/stockDiscoveryComponents/stockSearchResults/StockSearchResults.js
+++ b/src/components/stockDiscoveryComponents/stockSearchResults/StockSearchResults.js
@@ -37,6 +37,7 @@ function StockSearchResults({keyword}) {
         {loading ? <LoadingSpinner /> : error  ? <MessageAlert variant='danger'>{error}</MessageAlert> :
         Object.keys(stocks).length !== 0  ?
         <>
+        <div className="stocksearchresults">
         <h3 className="stockdiscoveryRow">Showing Results for "{keyword.trim()}"</h3>
         <Container>
         <Row md={1} xs={1}>
@@ -49,12 +50,14 @@ function StockSearchResults({keyword}) {
             </SideScrollMenu>
          
          </Row>
-         </Container>       
+         </Container>     
+         </div>  
         </> :
     
         <>
         <MessageAlert variant='danger'>No results match your search term "{keyword}"</MessageAlert>
         </>
+      
          
         }</>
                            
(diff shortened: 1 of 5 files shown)


src/components/stockDiscoveryComponents/stockSearchResults/StockSearchResults.js
```diff
@@ -37,8 +37,9 @@ function StockSearchResults({keyword}) {
         {loading ? <LoadingSpinner /> : error  ? <MessageAlert variant='danger'>{error}</MessageAlert> :
         Object.keys(stocks).length !== 0  ?
         <>
+        <div className="stocksearchresults">
```
WarrenK-design, line 40: Again have a look into Flexbox using containers, I see you have replaced Container with a div here and styling has effected the search bar. 

## Styling as Div 
When code is ```div className="stocksearchresults"```
image

## Styling as Container 
When code is just ```<Container> ``` no class. You can use ```<div>``` if you want but just might be handier if bootsrap fixes these issues for you. 
image

WarrenK-design (changes requested):
Hi Becks, 

Looks good though I think if we aren't changing the background colour of the cards or the screen itself the border needs to be more defined on the cards. 

image

Is a little overwhelming on some screens but ah that's only small thing.  
image

Netflix get away with using images for borders as backgrounds but images bit tricky and time consuming, think the black border will do. 
image

```
.tickercardstyle {
  width: 10rem;
  height: 100%;
  display: block;
  border-width: 0.175rem;
  border-color: black;
}
```

Also getting this warning. 
image

beckskelly:
Cheers for the review @WarrenK-design especially over the weekend. I think the black is a bit intense, but I'm sure there can be a workaround, maybe a shadow or something. I know what you mean about the container, I didn't use one on purpose to get the div to fit the width of the screen, but I might give it a go just altering the container, as you pointed out, bootstrap works best with its own elements

beckskelly:
Ok so I made the following changes:

1. Added a container around the search results, fixing the styling issue on searchbar
2. Left the div as is for the summary page, containers don't like going 100% width
3. Made the border of the cards slightly fatter, darker, and added a shadow

Let me know what you think!
image

WarrenK-design (approved):
Hi Becks looks good, the shadows definitely better than the black border. 

Might be something to add to the spreadsheet but I dont mind doing it nearly have code but not priority but when I do a search for a large search like "a". This sidescroll is bit annoying./lot of white space. 

image


Id rather something like this . 
image

image

Again bit like Netflix 
image

joeycor92:
Looks great after these changes. Thanks very much for all of this!

PR #547, opened 2022-11-26
Sector Selection Buttons

WarrenK-design:
When a user clicks the select sectors button now, all the sectors will be deselected by default. 

image


DaraghK93 (approved):
Hi Warren, thanks for making this change. I think that makes more sense now. and is consistent with what the lecturers said on friday as well. good work

BearachB (approved):
Looks good @WarrenK-design! A bit clearer than before

PR #548, opened 2022-11-26
Game - Backend - Fixing Join league and League Schema

DaraghK93:
I've fixed the joinLeagueByCode controller to only allow you to join games that are not finished. added

Screenshot 2022-11-26 at 18 35 50

so it will only find games that are scheduled and active.

also fixed league Schema by making users, leagueAdmin, accessCode and portfolios required. This would have saved us time the last day had i done it originally, as the league wasn't saving the users Id, meaning the leagues weren't being presented on the screen. better to have this check here now so these errors don't turn to bugs down the line.

to test, send an access code of a league that's finished to 

http://localhost:3001/api/league/joinleague

{
    "accessCode": <code>
}

also works on the frontend, and error should come up for invalid access code.

### edit

I've now fixed a bug in createLeague as well. the endDates being sent down by the frontend were coming out differently than I expected, so the finish timeBased leagues trigger wasn't picking them up correctly.

I needed

Screenshot 2022-11-27 at 19 07 53

but was getting
Screenshot 2022-11-27 at 19 08 26

So now the league controller converts them:

Screenshot 2022-11-27 at 19 11 58

and sends them down like that

Screenshot 2022-11-27 at 19 12 08


joeycor92 (approved):
Important fixes. Seems to be working for me. Tried a few different things and couldn't break it. Great work.

WarrenK-design (changes requested):
Hi Daragh, 

Have noticed something when trying to create a league on the frontend now.  I have tried to create a value based league. 

The league didn't show up first of all which taught was weird because no error. Looking at the console though this was the return message. 

image

An errormessage is being returned but the status code is actually a 200 still when there is an error so the frontend thinks its ok, if you go to "Network" in google chrome dev tools. 
image
image

The express error causing the problem. 
image


For the response code fix add this in. 
image
Then on frontend should get this. 
image
Might maybe change ``` res.errormessage ``` it to nicer message like "Unfortunately we could not create the league at this time" or something for the user.

PR #549, opened 2022-11-28
fixing error when getting host and api key from parameter store

caolan-power:


DaraghK93 (approved):
thanks for this, hope it fixes it

PR #550, opened 2022-11-28
Backend - get portfolios/games for Redux

DaraghK93:
Query for getting the active portfolios for a user, to be stored in redux. Will be used in choosing a game/portfolio to trade with on the stock page. returns an array of objects in the form
[{
       "_id": "6383c384e1f95f5e41426c78",
       "portfolioName": "Stock Battle Royale Portfolio",
       "leagueId": "6383c384e1f95f5e41426c76",
       "leagueName": "Stock Battle Royale"
}]
where _id is the portfolio Id.

Send a get request, with a user token to

http://localhost:3001/api/portfolio/myGamesAndPortfolios

### DIFF
--- a/amplify/backend/function/expressServer/src/controllers/portfolioController.js
+++ b/amplify/backend/function/expressServer/src/controllers/portfolioController.js
@@ -293,10 +293,57 @@ catch (err) {
 }
 }
 
+const getMyGamesAndPortfolios = async (req,res,next) => {
+  try{
+    const userId = mongoose.Types.ObjectId(req.user.id)
 
+    const portfolios = await Portfolio.aggregate(
+      [
+        {
+          '$match': {
+            'userId': userId
+          }
+        }, {
+          '$lookup': {
+            'from': 'leagues', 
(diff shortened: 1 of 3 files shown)


amplify/backend/function/expressServer/src/controllers/portfolioController.js
```diff
@@ -395,20 +395,77 @@ const getLeaguePortfolio = async (req,res,next) => {
...
+        }
+      ])
+
+  // return the array of portfolios
+  res.json(portfolios)
+  }
+  catch (err) {
+    console.error(err.message);
+    res.errormessage = 'Server error';
+    return next(err);
```
WarrenK-design, line 461: Need to set status on this, same thing will happen as last route will return a 200

WarrenK-design (changes requested):

WarrenK-design (approved):
Looks good

caolan-power (approved):
Works as expected. LG2M R2M

PR #553, opened 2022-11-28
Game screen time line and value line

beckskelly:
Made 2x components for time based game and for a value based game. Will need to be configured properly when we get the data in but shouldn't be too bad.

Timeline
image

Value line
image


joeycor92 (changes requested):
Just a few comments on the value based first.

Could be an issue having the names on the bar, see below
image

Only other things are that it's a bit small on desktop and I'm not sure if it should be visible at all times, e.g. when I go onto the stocks section.

The other thing is that I don't seem to be able to see the TimeLine even when I've chosen a time based game, is that correct. It looks really good from the image but just not sure where I can actually see it.

beckskelly:
@joeycor92 thanks for the review!I thought about the names yeah, I can introduce a function that doesn't allow them if its below 20%, its a good idea. I don't have it calling any routes at the moment, to see the value based one change this line of code 
image

in "individualGameScreen.js"

Sorry that wasn't clear at all! I'm open to getting rid of it on some screens and not always having it, for example stock page. Will make that update and get back to you.

beckskelly:
OK, updated this, now game image and game name are updated in the top, as well as the time line now harasses you when there are less than 3 people in a game. It also now goes from timeline to value line based on game type
Think i got all the edge cases incase of a tie etc

image

image

joeycor92 (approved):
Again really good stuff!

Two comments. 

First is that there doesn't seem to be a check for if it the game is over. (Although think you added that while I was writing this).

Second is that the copy button should probably say copy when you hover over or have a finger pointer when you hover over. I knew what it was but it wasn't obvious. Again. Great stuff.

I'll approve for now. A fun thing to show off tomorrow!

WarrenK-design (approved):
Looks good Becks, adds to the Gamey feeling of it, really good work! 

Maybe increase font size of the access code under the title in next PR but looks good.

PR #555, opened 2022-11-28
Machine learning stock recommender system api integration2

BearachB:
# Changes Included in this PR
- Finally got the recommender system fully linked up to the live application. This took a huge amount of time to get right but it's finally there. Spent a lot of time trying to get the API call to work with Axios.
- Added Axios API call on backend to stockRoutesService.
- This is called in the stockController and saved as an array which is then fed into @DaraghK93's function to fetch them from the DB.

image

This includes changes to stockRoutesService.js, stockController.js, and stockSummary.js. For now the recommender system is displaying on the discovery page, and for now it's hardcoded to show recommendations for Disney (DIS), this will be updated tomorrow to take in the username, once the updated model is deployed.
image

There's a few commented out console logs that have been left in there to make updating tomorrow easier. These will be removing before the next (and hopefully final) PR. I also had to update the name of this branch as the original branch had commits from one of Caolán's branches that I had pulled down. I know there's ways to fix this but I just didn't want to spend hours trying to get that fixed when we need features out.

Note that I've used Axios for this so you'll need to install the deps.

# React console
image

# Next Steps:
- Update deployed version to take username as input.
- Update stockRoutesService to fetch username and input this into the getRecomms function.

### DIFF
--- a/amplify/backend/function/expressServer/src/services/stockRoutesServices.js
+++ b/amplify/backend/function/expressServer/src/services/stockRoutesServices.js
@@ -122,8 +122,21 @@ const getRecomms = async (stock) => {
 const getStockSummary =  (schema) => {
 const stocks =  schema.aggregate([
     { $facet: 
+    // agg query for recommended
+    {
+        recommended_stocks: [{ $match: { symbol: { $in: recommended } } }, {
+            $project: {
+                'symbol': 1,
+                'longname': 1, 'exchange': 1, 'logo': 1,
+                'daily_change.absoluteChange': 1,
+                'daily_change.percentageChange': 1,
+                'daily_change.currentprice': 1,
+                'esgrating.environment_score': 1
+            }
+        }],
+        
         // agg query for top environment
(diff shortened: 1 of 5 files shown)


amplify/backend/function/expressServer/src/services/stockRoutesServices.js
```diff
@@ -89,11 +90,43 @@ function getStockPriceData (stocks) {
 
 };
 
-const getStockSummary =  (schema) => {
+// This function makes an API call to the API gateway for the recommender lambda function
+const getRecomms = async (stock) => {
+    var axios = require('axios').default;
```
WarrenK-design, line 95: Each time this function runs you will keep requiring it again, this function is run anytime user visits stock discovery, I'd require it up the top of the file globally so only executes once.

BearachB: Good spot apologies for that, I noticed this last night and meant to change it but it skipped my mind. Will change this now.

caolan-power (approved):
Working as expected Well done

WarrenK-design (approved):
Looks Good!

PR #557, opened 2022-11-29
Backend error messages update

DaraghK93:
We weren't setting the status correctly in a lot of controllers so added in

res.status(500)

to a lot of them. also tidied up some error messages.

also fixed error middleware as per Warren's comments:

Screenshot 2022-11-29 at 13 55 02

Still, the errors aren't very descriptive, so Caolan and myself are going to go through that at some point. for the below. .derk isn't a domain name so the mongoose validator catches it. we should be telling people what exactly is wrong with the email, as from the below they wouldn't know why. 

Screenshot 2022-11-29 at 13 55 59


joeycor92 (approved):
Think you've covered them all here. They seem to be detailed enough. Maybe the two Tweet ones could be a bit clearer as it's just get tweets and get all tweets but I'm happy to approve.

caolan-power (approved):
Nice one yeah will be good to get better errors showing on the front-end

Better to prioritise this first and we'll get the rest after this sprint maybe

PR #559, opened 2022-11-29
Backend cancel limit order

joeycor92:
Hi Folks, 

Back at it again with the limit order routes. This time we are cancelling an order. 

The route is http://localhost:3001/api/portfolio/cancelLimitOrder for both sell and buy.

Simply create a limit order the way we've done it before and then get the transaction id and portfolio id for this
{
    "transactionId": "6384d4d1f828bd870250f4b6",
    "portfolioId": "6384d41bb620745f0835267c"
    }

For buy order it will decrease the frozenBalance and increase the remainder.
For sell order it will just decrease the frozenUnitHoldings.

The transaction will also say "CANCELLED" now so that it won't be taken into consideration again.


DaraghK93 (changes requested):
Hey Joey, nearly there. just noticed that if you put in an id that's a valid Id but doesn't relate to a transaction it infinitely sends the request.

Screenshot 2022-11-29 at 16 41 45

as it doesn't return an array that should just be if portfolio === null

caolan-power (approved):
Hi Joey wasn't able to break it so nothing to add other than Daragh's comment. Will approve now since you're making those changes

DaraghK93 (approved):
Thanks for the change. great work

PR #565, opened 2022-11-29
Leader board

beckskelly:
Most of the logic out for the leader board, open to changes
Essentially this component takes in dummy data, which will come in sorted for the leaderboard (this is correct @DaraghK93 ? if not I can use my sorting function), splits it into 2 arrays, uses the first array for the top 3 at the very top, and the rest is mapped in the table. Did my best to have it looking good for desktop for a change as well, let me know how effective it is.

Had a go putting in lots of color, think it livens the app up a bit.

Future work
Have to put in the logic to check if the daily change is positive or negative and then insert a red arrow if negative, wont be too difficult, just holding off as i have to calculate the this change on the frontend anyway as per conversation with @DaraghK93 .
I'll do this when the route is in place

image


joeycor92 (approved):
I think if you fix that username that ye mentioned in the chat, this is good to go.

We also discussed in chat the possibility of an image background, which could be very nice. 

Just wondering how we're ranking the players if two players are on the same score. Might be something to get Daragh involved with.

joeycor92:
> I think if you fix that username that ye mentioned in the chat, this is good to go.
> 
> We also discussed in chat the possibility of an image background, which could be very nice.
> 
> Just wondering how we're ranking the players if two players are on the same score. Might be something to get Daragh involved with.

Can see you just fixed the navbar thing actually. Well done

PR #566, opened 2022-11-30
Splash page

caolan-power:
Creating this PR to get eyes on it and I haven't been on front-end much at all so I need to make sure what I'm doing is right before I continue. 
Screenshot 2022-11-30 at 13 04 19

Screenshot 2022-11-30 at 12 51 43


beckskelly (approved):
Great start Caolan, happy to approve for now for you to build on

BearachB (approved):
Looks good @caolan-power, nice starting point to build on from here! Nothing really to add yet, if I think of anything else will get onto you!

PR #570, opened 2022-11-30
Leader board connect backend

beckskelly:
Connected backend to leaderboard, shows top 3 if there are more than 3 players in the league, otherwise shows the table. Thanks to @WarrenK-design for the help with this!

I'll fix change when valuehistory is properly coming in but leaving it as dummy data for now
One player is a bit sad looking but not sure what can do about that?

imageimage

image


joeycor92:
> Grand @joeycor92 I'll make them bigger, I can't do anything about the scroll with 1 person as without it the background just cuts off and it looks awful

That's grand, if it causes issues that we'll just ignore!

joeycor92 (approved):
Happy to approve now.

WarrenK-design (changes requested):
Hi Becks, 

This is great work! Only one tiny change but looks really good, leaderboard seems to update when I buy stocks/join games quite cool. 

On larger screens there is fair bit of blue but ah to be honest its an extra wide monitor looks great on normal desktop/mobile, don't change this just maybe something we may look at when fixing stuff up at end. 
image

beckskelly:
@WarrenK-design , thanks for the catch and additional reading, you've a good understanding of these requests which is great. changed that there let me know if its good!

WarrenK-design (approved):
Looks good Becks!

PR #571, opened 2022-11-30
Quickfix to hardcoded API call

BearachB:
This is a quick fix to address the discovery page that was broken by the new recommender system deployment.

The problem - Page not loading because no value is returned when the API call is made. This was a necessary result of updating the lambda deployment:

image

The fix - Updated the API to a new hardcoded value for a user account. This will then be updated and parametrised but I wanted to get something displaying tonight before logging off.

image


WarrenK-design (approved):
Looks good, hopefully temp fix works!

DaraghK93 (approved):
looks good. good thinking for the time being

caolan-power (approved):
LG2M

PR #572, opened 2022-11-30
Backend update routes for triggers

joeycor92:
Quick PR, same as the last one through Postman.

Will talk through it on a call tomorrow so that you both know what you have to test.

Shouldn't be too bad but lots of checks will need to be done.

### DIFF
--- a/amplify/backend/function/expressServer/src/models/transactions.model.js
+++ b/amplify/backend/function/expressServer/src/models/transactions.model.js
@@ -5,6 +5,7 @@ const TransactionSchema = new mongoose.Schema(
   {
     portfolioId: {type: Schema.Types.ObjectId, ref: 'portfolioData', required: true},
     stockId: {type: Schema.Types.ObjectId, ref: 'stock', required: true},
+    holdings: {type: Schema.Types.ObjectId, ref: 'holdings', required: true},
     units: {type: Number, trim: true, required: true},
     value: {type: Number, trim: true, required: true},
     date: {type: Date, trim: true, required: true},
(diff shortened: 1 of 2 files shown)


joeycor92:
Updated now, lots of testing required.

The main order of things would be:
Create a league.
Buy Limit Order to check that a holdings with 0 units has been created.
Buy to increase units.
Sell limit order to check that it's working.
Cancel sell limit order to check that it's working.
Sell market all stocks to make sure that the holding is still there when the buy limit order still exists.
Cancel buy limit order to make sure that the holding has been deleted.

DaraghK93 (changes requested):
Hey Joey, testing away here, all seems to work with regards to the holdings being created/deleted. However i just noticed that my remainder didn't drop when I did a limit sell. this should drop by the tradingFee amount yeah?

DaraghK93 (changes requested):
Hi Joey, thanks for that change. just noticed I was getting a server error when it was checking whether it should delete a holdings object because of .length instead of checking for null after a findOne

Screenshot 2022-12-01 at 15 38 44

just change to otherTransactions === null in both

You might have a quick look to see if you've that after any other findOnes

caolan-power (changes requested):
Hi Joey I'm getting an error when I go to market sell the last of my shares. 
if i do market buy market sell for the same amount it fails. but i can do 2x market buy and 1x market sell. If i buy 10 i can sell 9 so it must something something to do with the 0 check

caolan-power:
Great ! was able to sell the last of the stock

DaraghK93 (approved):
Nice one Joey, thanks a million for this. think I've checked everything and it seems to be working

PR #575, opened 2022-12-01
Trade page portfolio balance fix

WarrenK-design:
This PR deals with refactoring mainly the quantity select and portfolio balance components to get them to show correctly for buy and sell. Had to refactor fair bit of the state so just kept this PR to fixing them components to avoid it getting too big. 

# Quantity Select
Should be able to select NaN anymore, reuses the range slider component now. When you select "Sell" then it should no longer decrease the "Spending Power" it should now increase it. 
image
image

#  Balance Component 
No longer shows you losing money when you are selling, updated some of the colours on it. 
## Buy Order 
Red = Spending power going down as you are buying. 
image

## Sell Order
Green = Spending Power Going up as you are selling. 
image

# Order Summary 
The order summary should be a bit clearer but if you want me to change wording let me know. 
image

image

# Future Work
1. When Order is confirmed the trade page should update the balnces in the individual components on the trade page. 
2. Limit order has not been touched yet. 
3. The Buy/Sell buttons, a temporary fix for is in this PR for disabling  Sell if you don't have at least 1 share in that stock. These need to be broken out into there own component and should really be toggle buttons. May look a bit confusing now and actually looks like sell is enabled but just temp fix. 
image
4. Info buttons to explain things, for orders, trade fees, formulas for balances etc. Will make it easier for the user.

### DIFF
--- a/src/components/confirmOrderComponents/AreYouSure.js
+++ b/src/components/confirmOrderComponents/AreYouSure.js
@@ -7,7 +7,7 @@ import { API } from "aws-amplify";
 import MessageAlert from "../widgets/MessageAlert/MessageAlert";
 import LoadingSpinner from "../widgets/LoadingSpinner/LoadingSpinner";
 
-function AreYouSure({showState,setShowState,portfolioId,stockId,buyOrSell,orderType,newPortfolioBalance,amountSelected,qty,limitPrice,stockName}){
+function AreYouSure({showState,setShowState,portfolioId,stockId,buyOrSell,orderType,newPortfolioBalance,dollarAmountSelected,qty,limitPrice,stockName,gameTradeFee}){
     
     const [loading, setLoading] = useState(false);
     const [error, setError] = useState("");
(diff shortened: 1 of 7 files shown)


joeycor92 (changes requested):
Looks really good. Loving the colour additions to the balances. 

As mentioned, the names could be better, we discussed and said this could be the best way to go
image

Maybe say Change in Spending Power as discussed.

beckskelly (approved):
Hey @WarrenK-design , great work, going to approve, just 2x small things, just for the future, when you hit confirm order I feel you should get redirected, and maybe make that disabled button grey colored for now?

WarrenK-design:
> Hey @WarrenK-design , great work, going to approve, just 2x small things, just for the future, when you hit confirm order I feel you should get redirected, and maybe make that disabled button grey colored for now?

Yeah maybe a redirect but some success meessage somewhere I think needs to be there just to acknowledge was successful alright, maybe to the Leaderboard page? They will see standing change then

PR #577, opened 2022-12-01
Machine learning stock recommendation system backend integration update

BearachB:
# Changes
- Updated backend to pass in userID to the recommender API call. 
- Had to make getAllStocks route protected to do this.
- This can't be tested locally because the API is set to connect to the live database, so if you buy stocks locally it will update your local DB and not the live one where the account data is being read from.
- Once live, will be tested on the application to double check it actually works ahead of tomorrow's demo.

## React Console
image


caolan-power (approved):
Looks good to me recommendations coming through be good to test it on live site when deployed

DaraghK93 (approved):
returning from correctly from the API. looks good. very nearly there!

PR #578, opened 2022-12-01
More education data

joeycor92:
This is going to be a continuously updated page. 

I've added some videos here which cause an error when you click play. This seems to be an issue that can't be fixed which is mildly really infuriating. 

Have a read through, see if there's anything mad. I'll continue to work on it just to get it in a good condition before the big launch next week.

I had a bit of trouble with the sizing of the YouTube video so please let me know how I might fix.
image

Thanks


joeycor92:
Also code for responsive YouTube embed came from here in case ye are wondering why the specific numbers.
https://avexdesigns.com/blog/responsive-youtube-embed

beckskelly (approved):
Looks fantastic @joeycor92 ! fair play on the embedding, does NOT sound fun

We can look into resizing those videos smaller down the line

WarrenK-design (approved):
Hi Joey, 

Looks really good thanks for the work on this. 

I see the error with youtube vid yeah some CORS error but plays fine, its probably just some header or something we need to set in request. 

Just too right before we add anymore could we refactor it a little bit maybe because I can see it getting out of hand if it gets bigger. 

Each element in it for the most part is only made of these components and then they are copied and pasted down. 
```
<Accordion.Item eventKey=KEY>
<Accordion.Header><span className="semibolded">HEADING</span></Accordion.Header>
<Accordion.Body>
               AN EXPLANATION
 </Accordion.Body>
</Accordion.Item>
```

We could make another file ```lessons.js``` (or maybe multiple files for individual lessons) which export JavaScript objects containing the information/JSX we need to fill these lessons. 

Your ```collapsibleTable.js``` would just be a for loop then and might be more maintainable to add too. 

Good work though!

PR #579, opened 2022-12-01
Limit orders trigger - update holdings 

DaraghK93:
just adding the code for the update of holdings collection via the trigger for a limit buy. This is the first stage really but took a while to get here, so the rest should be fine. 

Just uploading the query to the repo to have it. let me know if you've any questions on it.


caolan-power (approved):
Saw this working on call so can approve straight away

PR #582, opened 2022-12-02
Game page update

beckskelly:
Weekend so no need to review but I'd appreciate a glance over it and suggestions so we can improve this to a good standard next week. Changes made as follows:

1. Made that copy function into a component, it changes into a tick when clicked and a little message comes up "Copied". This lasts for 5 seconds and then it resets. Left it as standard styling for now, changing the background to a lighter color as your suggestion @WarrenK-design seemed a little complicated for that component for one use case, so I left it for now.

image
image

2. I took your game summary component @WarrenK-design and threw it in for the "rules" screen. I wasn't sure what to do with "ESG restrictions", any suggestions would be good. Remember that I'm getting all the info from the league object, so should I put in the min esg rating for each E,S and G rating?
image

3. The leaderboard was problematic in the mobile view. We have to be practical about what we can fit on a tiny screen, so I figured out how to take away the avatar away for the mobile, and it updates to include it when you're bigger than 550px. What do people think about this? Is it small enough? I could change the title of the "Rank" column to just a # to save more space

Desktop view:
image

Mobile View:
image

4. I linked up the portfolio using the route provided by @DaraghK93, thanks for helping me with this as I was chasing my tail for a while. To view this component with dummy data look up /portfolio on the site. If you have no value history there's obviously no graph so it shows you this if your value history is empty: 
image

5. I improved the appearance of the holdings table in both mobile and desktop view. I centered everything, formatted etc etc, and then I changed it so now it shows more columns on the desktop nowand removed the expandable row. What do people think? is it enough?
Desktop view:
imageMobile view
image

6. If the user has no holdings, this message comes up:
image


beckskelly:
Update - correct message alerts added to no stock state for both components:

image

joeycor92:
Still going through but just don't want to forget as I'm going through.

Issue with some logos on iPad. Maybe a change to the media query.
image
and mobile
image


Getting a few errors here
image

and here
image

On the points bits above, only comments I have are that I quite liked seeing the avatars on the phone but if they will cause issues happy to see them gone.

Great work on the copied button, so good to have!!

For the portfolio page, other than the things mentioned above, I think it's great that you can see more info on the desktop it's a great addition!


This could probably be a link to the stock page maybe?
image

joeycor92 (changes requested):
As mentioned above, if you can make those changes. Thanks again for all the work over the weekend.

beckskelly:
Cheers @joeycor92 thanks for such a thorough review on a weekend!

1. updated the logos on the table to fix the sizing issues
2. Made the error messages links as discussed, they go to stock discovery for now but when @DaraghK93 has the game stock pages up and running can just redirect to them
3. fixed the sorting issue
4. sorted all those errors I think, I wasnt getting the one you were getting in the browser so maybe check again?
5. Put a background on the leaderboard

beckskelly:
Sound for the review @WarrenK-design , I change that up to this simple page: 

image

joeycor92:
Have the same error as Warren, not only for no holdings but when there are holdings as well.

Also I mentioned the issue with the navigation. Everything seems to be on the same page now when I have holdings so just an if statement in the wrong place I'd guess!

beckskelly:
Grand, have managed to get it working again, also upated the graph message to: 

image

joeycor92 (approved):
Think you have everything we asked for (and it was a lot)! I'm happy to approve now. I didn't come across any problems.

WarrenK-design (approved):
Hi Becks, 

Seems to be working now no errors! Sorry about all the changes requested!

Good Work!

PR #586, opened 2022-12-05
Buy sell buttons

WarrenK-design:
Refactored the buy/sell buttons to the original Figma/Input taken from Damiens comments. Also refactored the Order Type Buttons, made them stand out more which one is selected. Took the explanation from the education section for explaining the order type. 

Made the portfolio name a heading on the page too to make it clear what portfolio you are using. 

Damien mentioned too that having "Let's Trade" and then "Confirm Order" on next page was confusing, should use "Trade" consistently so changed button to "Review Trade!" on the bottom. In the pop up are you sure changed "Order" to "Trade" where it made sense too. 

# Mobile View
image

# Desktop
image


beckskelly (approved):
@WarrenK-design , I really like how this page is coming along!

For the future my only suggestion is to make the inner container thinner for desktop, like how I did for the leaderboard page. The contents do look their best on a thinner screen, its a bit spread out on desktop. 

I think as well down the line a few icons/ cartoony bits around the order types would be a bit of fun, add more gamey elements and make it more understandable if they were relevant.


Happy to approve for now if you want to just keep going with functionality and getting the main user evaluation comments out. Well done

WarrenK-design:
> @WarrenK-design , I really like how this page is coming along!
> 
> For the future my only suggestion is to make the inner container thinner for desktop, like how I did for the leaderboard page. The contents do look their best on a thinner screen, its a bit spread out on desktop.
> 
> I think as well down the line a few icons/ cartoony bits around the order types would be a bit of fun, add more gamey elements and make it more understandable if they were relevant.
> 
> Happy to approve for now if you want to just keep going with functionality and getting the main user evaluation comments out. Well done

Cheers Becks, 

Yeah all this in middle looks rotten alright may take what you have done on leaderboard and make it smaller on desktop. 
image
Maybe even some background or something then. 

Yeah a few more cartoony icons be good, the + and - icon on the buy/sell are straight taken from Revoluts interface not very fun.

joeycor92 (commented):
Think everything is working well. As discussed, only difference I'd make is to have a reference to the trading fee on the page. 

The only other thing, and maybe it's FW, is that I scrolled straight past the portfolio name and didn't even realise that was my chosen portfolio. It's good to have it there, just maybe we need to find a way to make it more obvious.
image

beckskelly:
Had another look again there, the new dropdown position is much better than before and it works well! Fair play @WarrenK-design , great work

joeycor92 (approved):
Great changes. That limit order slider seems so finnicky but I think you did as well as you could do without having to make the user specify one first before going onto another page. I think you've reduced the number of click well in that regard.

The refining of the dropdown really helps the user to know which portfolio they are buying for as well. This is a great addition. Fair play.

Overall it's looking really well. Fantastic work on it.

PR #587, opened 2022-12-05
Machine learning stock recommendation system api call update

BearachB:
# Update
- This minor update adjusts how the API call is constructed.
- This is done to line up with the new API request body format. This has not been deployed yet, as it will break the Discovery page. I will wait until this is approved to deploy the updated lambda, then merge this immediately after so as to minimise downtime.
- Old request body format: `{
    "userid": "ObjectId('6388e0104af97027c17140b5')"
}`
- New request body format: `{
    "userid": "6388e0104af97027c17140b5"
}`
- This needed to be done as I had a mistake in the previous Lambda function when querying the userID. It turns out you need a separate BSON module from PyMongo to do this.


WarrenK-design (approved):
Hi Bearach, 

Looks good hopefully works!

PR #588, opened 2022-12-05
Backend - Game - Game stocks route start

DaraghK93:
This is the backend route for calling the playable stocks for the game screen. pushing as is, so @caolan-power can take over as discussed.

The controller function is in stock controller and right now this just calls the stockService and passes in the sectors and min ratings so that that query filters for those playable stocks.

### what needs to be done
- write the all stocks query, this will just be a find() that passes a similar match statement than the one in the service.
- write the conditional that checks if i's finding "all" or "summary"
- hook this up to the frontend. will be very similar to the main stock discovery page
- maybe add in search functionality, if there's time
- make a protected route

to test, send a body like this:

{
    "sectors": ["Technology","Healthcare"],
    "minErating": 750,
    "minSRating": 0,
    "minGRating": 0
}

(can change these fields to test)

route: http://localhost:3001/api/stock/gameStocks/game

the game param will probably change to the keyword

### DIFF
--- a/amplify/backend/function/expressServer/src/controllers/stockController.js
+++ b/amplify/backend/function/expressServer/src/controllers/stockController.js
@@ -203,6 +203,12 @@ const getGameStocks = async (req, res, next) => {
     // there will be an if statement that checks req.query and if it's 
     // keyword = all it finds all, or keyword = summary it gets the summary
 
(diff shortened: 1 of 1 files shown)


caolan-power (approved):
Thanks for this Daragh. Set up now to take over thanks

PR #589, opened 2022-12-05
Updated sort order of recommendations

BearachB:
# Changes:
- This change has updated the sort order of the recommendations to preserve the original cosine similarity order from the recommender API.
- It turns out when you use the $in Mongoose operator, it sorts the resulting array alphabetically by default. This was corrected by changing the sort order using the method found in this SO post: https://stackoverflow.com/questions/22797768/does-mongodbs-in-clause-guarantee-order/42293303#42293303. This basically works by appending a new attribute to each result and then sorting by that.

## Screenshot of API call showing correct cosine similarity ordering.
image

## Screenshot of the order change showing the correct order:
image


caolan-power (approved):
Nice one works as per screenshot

DaraghK93 (approved):
unreal. very nice elegant solution too. well done

PR #591, opened 2022-12-05
Backend leave game

joeycor92:
I've created a leaveLeague route now.

Admins can't leave their own leagues.

First create a league with one player and then join with another using the code and the route
http://localhost:3001/api/league/joinLeague (alternatively you can actually do this through your local app).

Then use http://localhost:3001/api/league/leaveleague to leave with the body looking like this
{
    "portfolioId": "638e2c525218379a73ab1602",
    "leagueId": "638e17f3dc9cd614a17d7c05"
    }

Hopefully this is ok.

### DIFF
--- a/amplify/backend/function/expressServer/src/controllers/leagueController.js
+++ b/amplify/backend/function/expressServer/src/controllers/leagueController.js
@@ -575,6 +575,7 @@ const leaveLeague = async (req,res,next) => {
       )
     }
     const portfolio = await Portfolio.findOne({_id: req.body.portfolioId, leagueId: req.body.leagueId, userId: req.user.id}) 
+    // check if a portfolio with league ID and user ID in it
     if(portfolio === null){
       res.status(404)
       res.errormessage = 'No such portfolio exists.'
@@ -584,6 +585,7 @@ const leaveLeague = async (req,res,next) => {
         ),
       )
     }
+    // check if league exists with that portfolio and that user
     const league = await League.findOne({_id: req.body.leagueId, portfolios: req.body.portfolioId, users: req.user.id}) 
     if(league === null){
       res.status(404)
@@ -594,6 +596,7 @@ const leaveLeague = async (req,res,next) => {
         ),
       )
     }
+    // the league admin can't leave the league
     if(league.leagueAdmin === req.user.id){
       res.status(400)
       res.errormessage = 'The admin cannot leave the league.'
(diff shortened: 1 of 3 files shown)


DaraghK93 (changes requested):
Hey Joey, nice work. Just a few things I noticed. getting a server error for incorrect portfolio and league Ids so you need to do this check, like we did in the last few

 // check that the leagueId can be cast to valid objectId
    if (mongoose.Types.ObjectId.isValid(req.params.leagueId) === false ){
    // check that the stock ID is correct
    res.status(404)
    res.errormessage = 'No league found'
    return next(
      new Error(
        'No portfolio found'
      )
    )
  }

functionality all there though I think. will ruin a few more tests in the morning with a clearer head.

joeycor92:
> Hey Joey, nice work. Just a few things I noticed. getting a server error for incorrect portfolio and league Ids so you need to do this check, like we did in the last few
> 
> // check that the leagueId can be cast to valid objectId if (mongoose.Types.ObjectId.isValid(req.params.leagueId) === false ){ // check that the stock ID is correct res.status(404) res.errormessage = 'No league found' return next( new Error( 'No portfolio found' ) ) }
> 
> functionality all there though I think. will ruin a few more tests in the morning with a clearer head.

Thanks for those comments, have made those changes.

As per the call this morning we also have it that the user can't rejoin the league after they have left, as they could just keep restarting and could potentially win this way.

joeycor92:
Three changes.

1. Remove league from user instead of user from league.
2. Update an error message for join league if a user has left a league and is trying to rejoin.
3. Update for sell limit stock so that the user can't go negative. Looks like this
image

DaraghK93 (changes requested):
Hey Joey, it's mostly there but I'm having the issue that the league Admin can leave the league. it could be due to the change for rejoining

joeycor92:
Thanks Daragh, update made and now the admin can't leave the game.

caolan-power (approved):
Hi Joey.

Works as expected well done.

When I try to leave a league giving someone elses portfolio id it says no such portfolio exists. So all good and cannot rejoin so well done.

I did get an error when trying to click back in to a game that i had left on the front-end but i wasn't able to reproduce so im not sure what went wrong

it flashed the screen saying 'no portfolio found'. 

Anyway good work and it's cool to see this working too

joeycor92:
Thanks Caolán, I've actually just changed it so that users can rejoin. This seems to get rid of this issue. To get rid of the old games from a user's page there would need to be a lot of changes to the logic of other routes. There's no point breaking this for a small detail and can be considered future work. 

I've also updated so that they can only leave games that are not finished.

DaraghK93 (approved):
Hi Joey, great work. couldn't break it. pity about the difficulties in rejoining but I think this works for now. Thanks

PR #592, opened 2022-12-05
Game screen - branch not cursed?

beckskelly:
I made a new branch and checked all the new code again, fingers crossed this one isnt cursed.

Main change vs before is this 

hopefully this sorts our issues before and we can leave this branch! 
image


joeycor92 (approved):
This seems to be working for me anyway. Although it worked for me the last time as well. I'm not getting any errors at all.

WarrenK-design (approved):
Hi Becks, 

Good work seems to be working no errors. 

Just one think for FW, on the portfolio holdings dropdown I think if I have less then 5, 10, 15 etc I shouldn't see them took me a min and was bit confusing. Mabe just the show all or disable it if they have this little stocks? 1 being the smallest number and down bottom bit confusing too because others are ordered but I know this is the "all" button. 

image

beckskelly:
Cheers @WarrenK-design , i completely agree with that, I'll add it to the bugs spreadsheet

PR #593, opened 2022-12-05
backend - Limit orders - add to stock price lambda function

DaraghK93:
near the end of this limit order saga I think. I've added it to the end of the stockPrice intra day function and It's working for me locally when I mock the function. This is a bit of a big one to test, I've tried to detail it all below.

## best way to test

create a few limit orders for a user
make them varied enough. maybe create some that add to holdings that already exist and some that are just created

## How to run

1.  create a .env file in the root of the stockPriceIntraDay folder and add

ENVIRONMENT=dev
MONGOURI=<your URI>
STOCKCOLLECTION=stocks
HOLDINGSCOLLECTION=holdings
PORTFOLIOCOLLECTION=portfolios
TRANSACTIONCOLLECTION=transactions

2. cd into that folder and run "pipenv shell" 
3. run "amplify mock function"

## make sure it worked

ensure that for a limit buy that goes through:
the frozenBalance drops by the value of the order
the units of the transaction are added to the holdings units
The transaction is set to "completed"

ensure that for a limit sell:
the frozenUnitHoldings decreases by the transaction units
the remainder increases by the value of the transaction
The transaction is set to "completed"

## note

the logic is currently wrong, so that we can test it
it checks for limit buys where the price is **greater than** the limit price, so that we can make it work. same with the limit sells. it actions ones that are less than the limit price. so be sure the make the limitValue way lower than the actual price for a buy and way higher for a sell so we can be sure it will work.

This will be changed once we're sure the live one is working with that logic


joeycor92 (approved):
This was a tough one for you.

I've tested with a few. It seems to work with the backward logic. It should be that it will work when you've changed to lte and gte for buy and sell respectively.

Really clear instructions on everything that should happen as well.

Serious work. Well done. Nearly there on backend functionality I think.

PR #594, opened 2022-12-06
backend for game stock page. type=all or type=summary

caolan-power:
Small enough PR building on from Daragh's start.

Backend for the game stock page finished. The route takes a param of 'all' or 'summary'. 'summary' matches the stock discovery page without the recommended stocks and the 'all' returns all tradable stocks.

Also in 'all' includes the basic pagination. Im hard-coding the page size to 20 for now but can easily change this depending on front-end needs.

TO TEST

http://localhost:3001/api/stock/gameStocks/all
OR
http://localhost:3001/api/stock/gameStocks/summary

SAMPLE BODY

{
"sectors": ["Technology"],
"minErating": 0,
"minSRating": 0,
"minGRating": 0,
"pageNumber": 0
}

Will move onto front-end then after this

### DIFF
--- a/amplify/backend/function/expressServer/src/controllers/stockController.js
+++ b/amplify/backend/function/expressServer/src/controllers/stockController.js
@@ -191,13 +191,33 @@ const getStockBySymbol = async (req, res, next) => {
   }
 };
 
+// @route   GET api/stocks/gameStocks/game
+// @desc    Get stocks that are playable in a game
+// @access  Private - add auth middleware to make it private
 const getGameStocks = async (req, res, next) => {
   try{
-    // the game rules are sent in the request body
-    // this might need to be changed but is ok for now
-    // this gets the game summary
+    const type = req.params.type;
+    let gameStocks;
     const {sectors,minErating,minSRating,minGRating} = req.body
-    const gameStocks = await stockService.gameStockSummary(sectors,minErating,minSRating,minGRating)
+    
+    if (type === 'summary'){
+      // the game rules are sent in the request body
+      // this might need to be changed but is ok for now
+      // this gets the game summary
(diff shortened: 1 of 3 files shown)


joeycor92 (changes requested):
Seems to be working well.

You probably need to include some more error checking. 
E.g.
A game will only have one of E, S or G rating so should check that
I put in a random word in the sectors and it doesn't tell me specifically why no stocks were found, should say something like "Not a valid sector"
It won't affect much but the input for the ratings, like it shouldn't be negative or a string. It says "No stocks found for this game" but should probably say something about it not being a valid input

It seems to be working well so once those are done it should be ready to go

DaraghK93 (changes requested):
Thanks for this Caolan, works on the backend but agree with @joeycor92 on the error checking.

I think passing all E,S and G ratings in is fine because it will have to do that. by default they're 0, so it won't affect anything.

think you need a check to make sure they pass all the arguments firstly. then like Joey said, check that the sectors are correct and that E,S and G are numbers. that should do it

DaraghK93:
thanks for the changes Caolan. as discussed on discord, you'll need to change the forEach to a forOf

caolan-power:
Added a search stock route. Needed a new route to get avoid request issues. 

ROUTE: 

http://localhost:3001/api/stock/gameStocks/search/keyword

DaraghK93 (approved):
thanks for those changes Caolan. happy to approve. there might need to be a few modifications later on, when we fix the main search a bit. but important we get the frontend going now and all seems to be correct

joeycor92 (approved):
Thanks for making those changes. Hopefully that's enough error checking anyway. Happy to approve so you can get going on the front end.

PR #596, opened 2022-12-06
Backend - limimt order function - update

DaraghK93:
Hi lads, just updating the logic here. the opposite way worked so just changed put in

Screenshot 2022-12-06 at 17 38 06

Screenshot 2022-12-06 at 17 38 23

should be fairly obvious from the 'files changed' section anyway.

no real way to test this. we'll add a few limit orders for stocks that we think should fall below that and hopefully it will work.


joeycor92 (approved):
Yeah this makes sense.

Thanks

caolan-power (approved):
Looks good

PR #597, opened 2022-12-06
fix create league

DaraghK93:
I've just changed this slightly. I needed to set all hours on the endDate to 0 so the trigger can match it exactly to today. I was doing this before creating the league object so the object was overriding my

endDate.setHours(0,0,0,0)

and was taking in the hours sent by the front end. I've tested and this shouldn't change anything.


joeycor92 (approved):
Yeah that makes sense. Tested there by creating a league and seems to work.

caolan-power (approved):
Seems to work for me nice one

PR #598, opened 2022-12-07
Limit orders

WarrenK-design:
This PR lets you complete limit orders through the frontend. 

You set the price you want to for a unit of stock and then how many stocks you want for both buying/selling. 

image

For a limit sell I have changed the balance component to "Potential New Spending Power" instead of "Spending Power" as it is not certain that you will actually get this money. The info box for limit sell is also different for this component, goes through example using the figures on screen of how its calculated. 

image
image

There is also error handling in the components now which should disable the Trade button until the user fixes the error. 
image

## Warnings 
If you mess around with the slider you will get this warning, its not an error and wont break the page just a warning. The dependencies are added to each of the useEffect arrays so you shouldn't get a terminal warning in your vs code, it's just when you move the slider a lot it causes useEffect to be called in multiple places as everything is a calculation based off the slider numbers.    
image

## Future Work
1. Shouldn't have to reload to see updates when you confirm order. 
2. "See More" section of the order Summary as Paul mentioned. 
3. Redirect possibly to different page on successful order? 
4. Fix sizing of some components for Desktop view.

### DIFF
--- a/src/components/confirmOrderComponents/BuyOrSell.js
+++ b/src/components/confirmOrderComponents/BuyOrSell.js
@@ -39,20 +39,4 @@ function BuyOrSell({state,setter,holding}){
 
 }
 
-
-//   {choices.map((radio, idx) => (
-//           <ToggleButton
-//             key={idx}
-//             id={`button-${idx}`}
-//             type="radio"
-//             variant='outline-primary'
-//             name="radio"
-//             value={radio.value}
-//             checked={state === radio.value}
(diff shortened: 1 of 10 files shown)


joeycor92 (changes requested):
Hi Warren,

Great work, did a good few things to try break this.

**Review Trade Disabled**
As discussed, if I disable the review trade button on limit page like this
image

It stays disabled when I go back to market.
image

Other issue will have a backend fix but probably needs to be blocked on the front end as well.
image

PR #602, opened 2022-12-07
Jwt error

WarrenK-design:
Fix for error Paul ran into hopefully, think issue was that his JWT expired. 

If it is less than a day until the JWT expires now you will be forced to log back in. I have set the JWT expiry time to 2 days now instead of 4 so we can see if it works over next few days. 

Checks it in the ```verifyJWT``` function in ```userActions.js```, the app wont load at all until this check is done shows spinning circle in the ```App.js``` component while loading in the ```validJWT``` redux state is true, quite quick check though you'd sort of miss it. 
Hopefully fixes any issues.  

**NOTE** - Did need to install a new package do a npm install before running.


caolan-power (approved):
Lowered the expires in time and it forced re login on reload so works well. 

Nice one

DaraghK93 (approved):
that worked for me Warren, thanks. same as Caolan, lowered to a day and a minute and I was logged out then when I refreshed or re entered the url

PR #603, opened 2022-12-07
Backend route delete league

joeycor92:
Similar to leave league - this is delete league.

It doesn't actually delete the league but instead changes active to false and finished to true.

It checks that the person deleting the league is the admin and also that the league is not finished (it could be scheduled so active would be false).

I also utilised Daragh's finish league trigger to create a final league table as well. 

**Sample Input**
http://localhost:3001/api/league/deleteleague
{
    "leagueId": "6390b7833b0d61314ff08075"
    }


DaraghK93 (changes requested):
Hey Joey, just one thing straight away, you won't need this
Screenshot 2022-12-08 at 09 37 25

because the aggregate below has this
Screenshot 2022-12-08 at 09 37 36

joeycor92:
As discussed, the above wouldn't work as merge returns an empty array. So just returning a 200 status. Thanks for the feedback.

DaraghK93 (approved):
thanks for the changes. all good, well done

caolan-power (approved):
Thanks for helping me test it. Silly mistake after a long few weeks

Looks good well done

PR #604, opened 2022-12-07
fixed individual stock screen error for brk-b and bf-b

DaraghK93:
very quick one here. we were previously checking the href to get only capital letters and then query by stock symbol, however for brk-b and bf-b that was sending the request as 'brkb' and 'bfb' so they weren't being found.

changed to the useParams hook instead of parsing the href. just one line:
Screenshot 2022-12-07 at 18 01 39

and sends the request to 
Screenshot 2022-12-07 at 18 01 57

All running with no errors on my end. let me know if I missed anything here.


WarrenK-design (approved):
Looks good

BearachB (approved):
LG2M. Nice simple solution!

PR #607, opened 2022-12-08
User avatars on leaderboard

DaraghK93:
Ok so few things here:

1. fixed the user model to take out what we don't need
2. added the default avatar to the user so when they sign up it's just the dollar sign. they can change then in user settings
3. brought back the user id in login and register for Joey's delete game route
4. changed the leaderBoard frontEnd to show the avatar. if they don't have an avatar (which will be the case for any users in the DB made before this) it defaults to the dollar sign as well.

I think test by creating and joining leagues and making sure the avatars show. try changing it and make sure it shows, etc


joeycor92 (approved):
All seems to work. Checked my redux on login and id is being returned. 

Changed my avatar as well and the leaderboard is working as well.

I want more avatars to choose from.

beckskelly:
Looks great, thanks for this!

beckskelly (approved):

PR #608, opened 2022-12-08
Update limit order lambda for date

DaraghK93:
Just added the current date to all actioned limit orders

to test, create some limit orders on your local and maybe change the logic in the buy and sell query, ie gte -> lte and vice versa. can do it here
Screenshot 2022-12-08 at 18 39 07

and the same for sell if you want.  then amplify mock function and check your local database to see if they went through. mine did anyway, time came through nicely

Screenshot 2022-12-08 at 18 40 32


joeycor92 (approved):
Great stuff. Makes a lot more sense to have the date as the date completed.

PR #610, opened 2022-12-08
Transcation history table started

beckskelly:
Have the transaction table nearly done, a few cosmetic issues with the filtering and I need to implement pagination, but general functionally should be there. 
Cant sort by limit value as i had to calculate this in table

Filters by 
1. pending, cancelled, completed
2. Buy - sell - all

Also sort function as seen in previous table

Same expand function as before

image
image

Future work: 
1. pagination (need to sort logic in other table before lifting it over here)
2. Change navigation filter names to "pending limit orders" "cancelled limit orders"
3. Works for mobile and desktop but not for tablet, would need a little bit of work for all screen sizes
4. Cancel button currently does nothing


WarrenK-design:
Should I be getting market orders as pending? 
image
They are completed in the db
image

WarrenK-design (approved):
@beckskelly looks good, not to do with this pr but just to make note of that edge case for 0 dollar holding for next PR 
image

joeycor92 (approved):
Looks good other than the aforementioned error that you know about.

It looks good and the leaderboard seems to be working as well.

Great stuff.

PR #612, opened 2022-12-08
Frontend delete game leave game

joeycor92:
Just a quick PR on the leave game and delete game buttons,

Only league members who are not admin can leave a game so if you join another game you'll be able to see the leave button. It looks like this.
image

Only the league admin can delete a game. The page looks like this.
image

For both you'll get an Are You Sure component and then you'll have to return to the games page. There is also a dispatch in there so your redux should also update.

Let me know if any qs.


DaraghK93 (changes requested):
Hi Joey, well done, looks great and works well. just one thing, can you make it so that the delete league button disappears once the game actually finishes? also maybe that once a game is finished they can't leave it?

WarrenK-design (changes requested):
Hi Joey, 

Looks good works functionally, just as discussed on discord the "Delete League" button wording. 

Just too noticed something, and its an annoying one if I'm on the website when the admin deletes the league, I can still trade with a deleted leagues portfolio. 

League is over.
image
Went and made a trade, went through successfully can see the balance going down from trade fee.
image

It might be adding just a call to updating the active portfolios in redux on frontend for the individual stock page to try prevent it, I'll try have a think about it, I taught the backend check would of got this though can see active is false. 
image

joeycor92:
Thanks very much for the review Warren. I've made the following updates

### Updated to say cancel league and removed references to delete.
image

### Updated the trade routes to check that the league is active and if they've got that far it'll return this (feel free to suggest changes to wording)
image


### Updated that the individual stock page has a useEffect that will do a dispatch to update the available portfolios

Let me know if there are any changes

WarrenK-design (approved):
Looks Good!

DaraghK93 (approved):
great work Joey, works perfectly

PR #616, opened 2022-12-10
Fixing triggers

DaraghK93:
just updating the triggers, as there was a change due to a mistake in the logic for finishing the game. all working on live site currently


BearachB (approved):
As before, working on live site so can see it's working! Thanks for sorting this, handy to get the DB switched over today when things are relatively quiet development wise!

joeycor92 (approved):
Great stuff Daragh.

PR #621, opened 2022-12-10
Pre baked games

WarrenK-design:
The first page in the game creation flow now gives you an option to select a prebaked game. Four games for the moment one for time, value, ESG and sector based restrictions. When you select one of these it takes you straight to game summary screen. 

There still is the option to configure your own settings. 

It was a comment too that game creation doesn't look create on mobile, have started some of fix for this in this PR making the bottom buttons sticky on mobile. 

Mentioned on discord too, new folder under ```components``` called ```icons```, code for icons taken from this website  https://icones.js.org/ which is linked in the files, I did not write this code they are freely available SVG's. 

# Mobile View 
image
image

# Desktop 
image


beckskelly (approved):
Seems to work perfectly, thank you for this, I really like the new buttons as well for the mobile in both options and I like the little SVG's. Maybe we could add a few of these to the "summary" screen to liven it up at some stage?

joeycor92 (approved):
Actually looks so good! Great work. Have never seen someone set mobile screen the way you've done it before. 

The icons add such a fun element to it.

The one thing I'd say, and it's not really this PR at all, is just that the ESG restrictions page feels very long on mobile. It's not that important just wondering if there's anything we could do with it.
image

Happy to approve as long as you remove the console log.

PR #622, opened 2022-12-11
Backend daily data

joeycor92:
Daily data now coming through in this route.

It also checks if it's the weekend so that the graph won't be empty on Saturday and Sunday when no prices are coming through.

Send something like 
http://localhost:3001/api/stock/MMM

It'll return week, month, year, day in this kind of format.
image

I've included Becks as she's working on the front end and needs to do the manipulation on her side. If both sides are happy then I'll merge. While we're waiting for Becks to update, the graphs could be a little bit off but hopefully not for too long.

@beckskelly if we can keep the time that they're messed up down to a minimum that would be good.


caolan-power (approved):
@joeycor92 should the decimal be rounded like in the pic? im getting massive decimals for everything other than the last few entries ? maybe rounding has been implemented since my last db dump is it that it? Happy to approve though as the format is as it should

joeycor92:
> @joeycor92 should the decimal be rounded like in the pic? im getting massive decimals for everything other than the last few entries ? maybe rounding has been implemented since my last db dump is it that it? Happy to approve though as the format is as it should

Nice spot. Not an issue really. Basically the data that was brought in using yfinance api had all of the extra decimal places, so that's from whenever that was done. The rest of the data is coming from slickcharts which is only two decimal places. It's not a problem really but yeah the users might notice more detail on the older ones for 1 year.

beckskelly (approved):
Hey @joey, looks great, works perfectly on my end, I'm going to approve so I can get going on manipulating the format on my end in my own pr

PR #625, opened 2022-12-11
Frontend UI join game screen update

BearachB:
# Changes
- Added a new modal popup when a user successfully joins a league.
- Gives the user a link to their game homescreen or stock discovery. I tried putting in a link to the game they just signed up to however I could not get React to take in the league ID which is needed to construct the url. In the end rather than wasting more time on it I decided to just leave it as is. @DaraghK93 can attest to this, he helped me with some troubleshooting for a while until we decided to stop putting time into it.
- Confetti might be overkill but it was there on @WarrenK-design's original modal so I left it in.
- Also note that this is not in as a React component. I had it made as one at first however I was unable to pass the league name to the component for some reason, despite the same modal working when hardcoded onto the page. 

# Screenshots
## JoinGameSuccess Modal
image

## React Console
image

### DIFF
--- a/src/components/gameComponents/gameScreenComponents/JoinAGame/JoinAGame.js
+++ b/src/components/gameComponents/gameScreenComponents/JoinAGame/JoinAGame.js
@@ -100,7 +100,7 @@ function JoinAGame(){
                         <Modal.Body>
                             <Row xl={1} md={1} sm={1} xs={1} className="textCenter">
                         <Confetti numberOfPieces={500} recycle={false}/>
-                                <h2 className="greenSuccess"><GroupAddIcon /> Succesfully joined league: "{leagueName}"!</h2>
+                                <h2 className="greenSuccess"><GroupAddIcon /> Succesfully joined league: {leagueName}!</h2>
                                 <Col className="w-100 mb-4">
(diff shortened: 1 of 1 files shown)


joeycor92 (commented):
Great stuff, use of icons and stuff like that is perfect.

Just wondering would it make more sense to go to that game rather than all games?

Explore Stocks I'm ok with, although we should probably be encouraging them to go look at the stocks on the games page when that's up and running?

Only other thing, and it's nitpicking is that it's very close to the top on phone. Can we bring it closer to the middle? Like around where I have the pink box
image

joeycor92 (approved):
As discussed I've got that going to the other page now. I was just nitpicking with the positioning stuff so happy to approve that.

beckskelly (approved):
Looks great @BearachB I'll approve, I took the liberty of adding 2x more icons to the buttons, you can take out if you wish just thought it was more gamey

image

BearachB:
> Looks great @BearachB I'll approve, I took the liberty of adding 2x more icons to the buttons, you can take out if you wish just thought it was more gamey

Great addition @beckskelly! Thanks for that!

PR #627, opened 2022-12-11
Query for trade page

DaraghK93:
bringing back game rules: ESG minimums and Sectors

this is so Warren can put the check on the front end whether the stock they're buying is tradable within that game

updated the getLeaguePortfolio controller:
Screenshot 2022-12-11 at 18 51 05

coming in now when you're on that page

Screenshot 2022-12-11 at 18 47 24

Sample postman GET:

http://localhost:3001/api/portfolio/<leagueId>

empty body and send token


caolan-power (approved):
Works as expected Nice work

WarrenK-design (approved):
Looks good cheers for this Daragh!

PR #628, opened 2022-12-11
Show password and email registration fixes

caolan-power:
This contains fixes for username length validation on registration and change details, 

email validation now matches what is enforced on the mongo db level and the appropriate error returned which was already being handled. this is in registration and change details.

User settings error placement moved under the headings so hopefully more clear now. 

changeuserdetails form fields should clear after submitting the form.

show password option on registration screen and displaying the requirements.


caolan-power:
Added code to remove errors from redux state when a user reloads the page.

Dispatches the RESET_ERROR action

caolan-power:
removes errors when component unmounted for login register and change details pages

BearachB (changes requested):
# Working Fine
1. Username length validation.
2. Email validation.
3. Show password on registration (big fan of this).

#Changes Requested
1. Error Message should read "less than or equal to 12 characters" or "must be less than 13 characters" as it lets you input a password of length 12.
image

4. I get logged out and redirected to the splashpage when I click on user settings from the Navbar. I hope this isn't a problem on my end, but I tried in Chrome incognito and also in Firefox and same result. This is working on the live site, it's just on this branch that I can't.
image
image

I couldn't test the other functionality because it won't let me into user settings.

caolan-power:
Hi Bearach thanks for that I was doing something stupid referencing the same reducer when trying to clear the errors when you navigate away from or reload the page. Should be fixed now and will change that error message too. Well spotted

BearachB (approved):

BearachB:
OK that's all good I can change my details and it's all updating on the front end. Works as expected.

DaraghK93 (changes requested):
Hi Caolan, great work here. just two things I think could work
1. adding show password to login. this shouldn't be too bad
2. when you enter the form, the fields reset. this works perfectly. but if there's an error, it still does it. it might be an idea not to reset them after an error. as discussed, if this takes a substantial amount of time then leave it. it works as is.

DaraghK93 (approved):
looks good, thanks for that change

PR #629, opened 2022-12-11
Check for sectors and esg trade page

WarrenK-design:
Error for the sectors and ESG rating on trade page. 

Will show an error if the user cannot trade a stock due to the ESG rating not being met or the sector is not allowed, the confirm trade button will be disabled. 

image

image
image


beckskelly (approved):
works perfectly warren so I am going to approve, if you could indulge my ocd and change the classname of this link to "linkStyle" I'd be a very happy front end dev

image

DaraghK93 (approved):
happy with this, thanks for the changes

PR #630, opened 2022-12-12
sorting tweet by newest-oldest

caolan-power:
Tiny one to order tweets newest-oldest so we can see them when returning only 5 from the db


BearachB (approved):
Good stuff @caolan-power, nice little update. 
All working on my end. Tested it on a brand new DB dump to local and it's definitely sorting by most recent now!
image

PR #633, opened 2022-12-12
Trade page min stock purchase

WarrenK-design:
***NOTE*** maybe don't review this until Joey fixes the stockPrice error, it looks like I changed the service but just needed to change it a bit locally because of the server error.   

Min number of stocks you have t buy is now 1 unit, avoids it looking like you are purchasing 0 stocks, try NVR its an expensive stock. 

Price in my db is $4,524.00

image

Fixed too the selling of stocks so the units you own have to be greater than 0 to avoid having fractional shares stuck in portfolio. 

If you try a portfolio that doesn't have enough to buy 1 stock the error from before still appears. 
image


beckskelly (approved):
Works perfectly, thanks for this @WarrenK-design

joeycor92 (changes requested):
Looks good. Only thing, any chance of us fixing it so that someone can sell their stocks if the amount of the sale is more than the trade fee. Otherwise a user on 0 spending power is in trouble.

joeycor92 (approved):
Thanks for all those changes we discussed. 

Now I can finally get rid of Etsy!!!

Top notch as usual. Happy to approve and thanks for removing the console logs. 

Approved!

PR #634, opened 2022-12-12
Stock searching error

caolan-power:
this sanitises input on the stock search bar so it doesn't cause errors when searching for special characters. added '-' and '&' as we need them for searching i think. Any other symbols let me know


caolan-power:
new one up without formatter

PR #638, opened 2022-12-12
Create game date error

WarrenK-design:
Check for the time based game that the game as to be at least a game long. 

image

Also put border back on slider in this PR. 
image


beckskelly (approved):
Works perfectly @WarrenK-design , thanks for this, bit of work I imagine and makes the game more realistic!

DaraghK93 (approved):
thanks for this, a lot handier than having it there while they click back. good work

BearachB (approved):
Looks good @WarrenK-design. The game duration of 1 day warning works as expected.
The border around the number fields is a little bit ugly but does the job in letting users know they can type directly into the fields. Good stuff!

PR #641, opened 2022-12-13
fixing portfolio absolute and percentage change and totalvalue

DaraghK93:
ok so just fixed the values coming for the portfolio sections. if they have any holdings and have a day's worth of history, the graph will show and the totalValue of the portfolio will be shown at the top here

Screenshot 2022-12-13 at 18 11 02

that's on an active game. we've set the change as the difference between their current value and their starting value as per conversations with Becks to keep it consistent

on a finished game, it will be the same, however it gets the total value from the final standings.

Screenshot 2022-12-13 at 18 12 47

not ideal that we never really show the current value on the graph, or the final standings value but i think it makes more sense now like this


WarrenK-design (approved):
Looks good Daragh, cool seeing portfolio value

PR #642, opened 2022-12-13
nav logo

beckskelly:
added logo to navbar. there was side scrolling on splash page too so got rid of it

image


BearachB (approved):
Love it, definitely a small but important addition! Thanks @beckskelly

WarrenK-design (approved):

PR #644, opened 2022-12-14
Fix league values

WarrenK-design:
Fix for league values, parsed as float on frontend now


DaraghK93 (approved):
nothing like a last minute PR to get the heart rate going. well done for fixing

caolan-power (approved):
thanks a million

